PR #4, opened 2020-10-06
RFC006 Distributed Registry

reinkrul:


rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,146 @@
...
+
+TODO: What if 2 vendors have the same organization as client?
+TODO: Do we need start/end dates now we don't have organization certs anymore?
+TODO: Should 'name' by an array to allow aliases?
+
+#### Document type
+The document type for registering organizations MUST be `registry/organization`.
+
+#### Signer
+This document MUST be signed by the vendor identified by **vendorID**.
```
woutslakhorst, line 90: How to find the correct certificate and how does this work with past documents

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,146 @@
...
+
+### 4.4. Endpoint
+This document registers an endpoint for an organization which CAN be used to exchange data with that organization.
+
+#### Fields
+The following fields MUST be present:
+* **id**: MUST be a unique string identifying this endpoint.
+* **oid**: MUST contain the ID of the organization this endpoint belongs to.
+* **type**: MUST be a URN describing the type of this endpoint.
+* **location**: MUST be a URN pointing to where the endpoint is exposed (e.g. `https://nuts.nl/some/path`)
```
woutslakhorst, line 114: url?

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,146 @@
...
+* **Document**: a distributed document as specified by [RFC004](rfc004-distributed-document-format.md).
+* **Organization**: a care organization exchanging data with other care organizations over a Nuts network.
+* **Vendor**: an entity developing software for care organizations participating in a Nuts network. Organizations
+  access the Nuts network through their vendor's software.
+* **Endpoint**: typically a URI or URL exposed by a vendor which can be used by other vendors to pull data.
+* **Service**: a grouping of endpoints implementing a service required by a Bolt.
+
+Other terminology comes from the [RFC001 Nuts Start Architecture](rfc001-nuts-start-architecture.md#nuts-start-architecture).
+
+## 3. Updating
```
woutslakhorst, line 45: Does an update make a previous document obsolete? How does this work with timelined data?

reinkrul: Can you give an example?

rfc/rfc006-distributed-registry.md
reinkrul: URN can contain more types like telephone numbers or e-mail addresses

rfc/rfc006-distributed-registry.md
reinkrul: Signing certificate is present in the document (RFC004) so it can be verified against the truststore filled with all vendors certificates. What do you mean by past documents?

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,146 @@
...
+The following fields MUST be present:
+* **certs**: MUST be an array containing base64 encoded vendor CA certificates. There MUST be at least one entry.
+* **proof**: MUST contain cryptographic a proof that vouches for the vendor being a real software company. See section 5
+  for supported proof types.
+
+#### Document type
+The document type for registering vendors MUST be `registry/vendor`.
+
+#### Signer
+This document MUST be signed by the vendor itself.
```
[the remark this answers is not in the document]
reinkrul: Nope this is already supported by RFC004

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,146 @@
...
+
+#### Signer
+This document MUST be signed by the vendor itself.   
+
+### 4.2. Organization
+This document registers a care organization as a client of registered vendor (section 4.1).
+
+#### Fields
+The following fields MUST be present:
+* **id**: MUST be the organization's ID.
```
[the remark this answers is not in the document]
reinkrul: Added it inline for now, we might need to add it to separate RFC or Start Architecture (like the Identifiers page we have on readthedocs)

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,146 @@
...
+* **Organization**: a care organization exchanging data with other care organizations over a Nuts network.
+* **Vendor**: an entity developing software for care organizations participating in a Nuts network. Organizations
+  access the Nuts network through their vendor's software.
+* **Endpoint**: typically a URI or URL exposed by a vendor which can be used by other vendors to pull data.
+* **Service**: a grouping of endpoints implementing a service required by a Bolt.
+
+Other terminology comes from the [RFC001 Nuts Start Architecture](rfc001-nuts-start-architecture.md#nuts-start-architecture).
+
+## 3. Updating
+To update an object (vendor, organization, endpoint or service) the vendor MUST create a new document with the same identifier
```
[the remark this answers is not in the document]
reinkrul: Vendor identifier should be read from certificate since otherwise it's duplicate (both in certificate and in JSON). That leads to more validation logic (making sure they're the same).

rfc/rfc006-distributed-registry.md
woutslakhorst: What happens when you change the validity period of an endpoint? Is the existing record updated with a new record that adds an end_date plus a new record that has a start_date starting at the previous end_date? That last record would then not be an update just a new record.

rfc/rfc006-distributed-registry.md
woutslakhorst: If you say "MUST be signed by the vendor", you probably mean: "MUST be signed by the certificate currently in the truststore where the SAN matches the vendorID field...

If a Vendor CA gets a rekey, the current CA doesn't sign the older documents. If a new node replays the entire history, it'll have to replay it against older CA's.

rfc/rfc006-distributed-registry.md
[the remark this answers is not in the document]
woutslakhorst: Yes, but it needs to be very explicit: the vendor document doesn't have an identifier field but it does have an extra validation based upon certificate contents.

rfc/rfc006-distributed-registry.md
```diff
@@ -58,7 +58,7 @@ All certificates for a vendor MUST contain the same vendor ID.
 #### Fields
 The following fields MUST be present:
 * **certs**: MUST be an array containing base64 encoded vendor CA certificates. There MUST be at least one entry.
-* **proof**: MUST contain cryptographic a proof that vouches for the vendor being a real software company. See section 5
+* **proof**: MUST contain a cryptographic a proof asserting the vendor being a real software company. See section 5
```
woutslakhorst, line 61: still an a too many

rfc/rfc006-distributed-registry.md
```diff
@@ -67,15 +67,23 @@ The document type for registering vendors MUST be `registry/vendor`.
 #### Signer
 This document MUST be signed by the vendor itself.   
 
+#### Example
+```json
+{
+    "certs": ["base64-encoded cert here"],
```
woutslakhorst, line 73: Multiple CA's?

rfc/rfc006-distributed-registry.md
```diff
@@ -67,15 +67,23 @@ The document type for registering vendors MUST be `registry/vendor`.
...
+}
+```
+
 ### 4.2. Organization
 This document registers a care organization as a client of registered vendor (section 4.1).
 
 #### Fields
 The following fields MUST be present:
-* **id**: MUST be the organization's ID.
+* **id**: MUST be the organization's ID (AGB-code) encoded as URN, e.g.: `urn:oid:2.16.840.1.113883.2.4.6.1:1234` (where `1234` is the AGB-code).
```
woutslakhorst, line 83: Could you make the agb-code an example of a valid urn?

rfc/rfc006-distributed-registry.md
```diff
@@ -67,15 +67,23 @@ The document type for registering vendors MUST be `registry/vendor`.
...
 * **vid**: (vendor ID) MUST be the organization's software vendor's ID as string.
 * **name**: MUST be the organization's name.
-* **proof**: MUST contain cryptographic a proof that asserts the organization as being an actual care organization.
+* **proof**: MUST contain a cryptographic a proof asserting the organization as being an actual care organization.
```
woutslakhorst, line 86: so many aaaa's

woutslakhorst (changes requested):
I think a *validations* section would be nice for the different document types. A general validation for all document stating that required fields MUST be checked but also an additional validation section per document stating what extra things to check: a vendor update document MUST check if the VendorID in the SAN matches, or what happens when there are 2 Vendor documents with different keys?

rfc/rfc006-distributed-registry.md
reinkrul: In the current state (complete replacement) you'd indeed create a new record with the same ID and properties but with updated validity, which effectively replaces the endpoint. 

IMO updates through complete object replacement is the easiest way. The alternative is allowing incremental updates to JSON documents instead of replacement. In that case we should use a standard like http://jsonpatch.com/).

It all matters on the requirements of the system:

**Data size**
Do we care about bandwidth/storage? Incremental updates lead to less data.

**Simplicity**
How important do we find simplicity? Object replacement is simpler.

**Domain modelling**
Do we foresee somebody would want to fire domain-specific events when a particular property changes? This is easy with incremental updates but very hard/impossible with complete replacement because you'd have to derive the domain-specific event by checking deltas of object states. For example a change of `Endpoint.StartDate` can be mapped to an `EndpointValidityUpdatedEvent` but what if multiple properties change in the same delta? Fire 2 events but in what order?

(this is an actual problem I encountered in a system I built)

rfc/rfc006-distributed-registry.md
reinkrul: fixed :-)

rfc/rfc006-distributed-registry.md
woutslakhorst: Summary of discussion: separate distribution model of internal model. The distribution model can work with jsonpatch, but the internal model should store versions and when a specific version was valid.

rfc/rfc006-distributed-registry.md
reinkrul: There can be multiple entries

rfc/rfc006-distributed-registry.md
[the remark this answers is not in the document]
reinkrul: Added validation section

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+`test` operation to make sure the intended values are removed.
+
+## 4. Object types
+This section describes each of the supported object types.
+
+### 4.1. Vendor
+This document registers a vendor on the registry. Since new vendors can't connect to other nodes in the network since their
+CA certificate isn't trusted yet, another vendor SHOULD register the new vendor through the following process:
+
+1. Vendor *A* sends vendor *B*:
```
woutslakhorst, line 85: Isn't 1 irrelevant since A will create and sign the document?

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+```json
+{
+  "patch": [
+    { "op": "add", "path": "/prfs", "value": "some new proof" },
+    { "op": "replace", "path": "/crts", "value": ["CA certificate as base64 ASN.1"] }
+  ]
+}
+``` 
+
+Vendors SHOULD use `replace` to atomically revoke the lost certificate and add the new one.
```
woutslakhorst, line 157: maybe call it _replace in the current active truststore_ since old documents are still validated against the old CA. Also revocation is not a concept within the current setup

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+
+#### CA certificate private key leakage
+In case the vendor leaks its CA private key (or it gets stolen), it MUST revoke the associated certificate as soon as possible.
```
woutslakhorst, line 160: revoke == replace

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+```json
+{
+  "patch": [
+    { "op": "add", "path": "/prfs", "value": "some new proof" },
+    { "op": "replace", "path": "/crts", "value": ["CA certificate as base64 ASN.1"] }
+  ]
+}
+``` 
+
+If a vendor removes a certificate it SHALL NOT be re-added
```
woutslakhorst, line 172: After a rekey, it may not reuse the old key

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+TODO: What if the attacker re-adds the certificate after the vendor removed it, so it can still sign
```
woutslakhorst, line 173: not possible when old key is no longer valid (on denylist)

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+    { "op": "add", "path": "/prfs", "value": "some new proof" },
+    { "op": "replace", "path": "/crts", "value": ["CA certificate as base64 ASN.1"] }
+  ]
+}
+``` 
+
+If a vendor removes a certificate it SHALL NOT be re-added
+TODO: What if the attacker re-adds the certificate after the vendor removed it, so it can still sign 
+TODO: What if the attacker keeps adding anti-dated (timestamp in history) changes, branching from an earlier document before the certificate was revoked?
+TODO: Internal revocation list SHOULD be build when a certificate is removed
```
woutslakhorst, line 175: it'll be removed from the truststore in the OS ref impl. Maybe call it _appropriate measures against usage_?

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+  "patch": [
+    { "op": "add", "path": "/prfs", "value": "some new proof" },
+    { "op": "replace", "path": "/crts", "value": ["CA certificate as base64 ASN.1"] }
+  ]
+}
+``` 
+
+If a vendor removes a certificate it SHALL NOT be re-added
+TODO: What if the attacker re-adds the certificate after the vendor removed it, so it can still sign 
+TODO: What if the attacker keeps adding anti-dated (timestamp in history) changes, branching from an earlier document before the certificate was revoked?
```
woutslakhorst, line 174: This is where the max age of a newly received doc could help. This also determines an upperbound to the clockskey/downtime => CRL/OCSP caching averages => 3 to 4 days!

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+This document registers a care organization as a client of registered vendor (section 4.1).
+
+#### Fields
+The following fields MUST be present:
+* **id**: MUST contain the organization's ID encoded as URN. This can be any valid URN, for instance an AGB-code or
+  KVK-number e.g.: `urn:oid:2.16.840.1.113883.2.4.6.1:1234` (AGB-code). It is used to refer to the organization by the
+  vendor itself or other vendors. The organisation ID SHALL NOT be changed.
+* **vid**: (vendor ID) MUST be the organization's software vendor's ID as URN (e.g. `urn:oid:1.3.6.1.4.1.54851.4:1234`).
+  The vendor ID SHALL NOT be changed; relation to a care organization SHALL NOT be transfered to another vendor. 
+* **name**: MUST contain the organization's commonly known name as string. Only alphanumeric and space characters are allowed,
```
woutslakhorst, line 187: `-` symbol?

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+1. Assert all required fields are present.
+2. Assert all fields value formats are valid.
+3. For updates: assert only modifiable fields are updated.
+4. Assert the document is signed by the vendor identified by **vid**.
+5. Validate **prfs**:
+   * Assert there is at least one proof.
+   * Assert proofs are valid at the time they're introduced.
+   * Assert all proofs authenticate the same organisation.
+
+### 4.3. Service
```
woutslakhorst, line 218: Will have to be expanded with endpoint refs (uncertain how though)

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+This document registers a service provided by an organization and SHOULD describe the endpoints that SHOULD be used for a specified Bolt (use case).
+
+#### Fields
```
woutslakhorst, line 221: service must be care org specific

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+2. Assert all fields value formats are valid.
+3. For updates: assert only modifiable fields are updated.
+
+### 4.4. Endpoint
+This document registers an endpoint for an organization which CAN be used to exchange data with that organization.
+
+#### Fields
+The following fields MUST be present:
+* **id**: MUST be a unique string identifying this endpoint.
+* **oid**: MUST contain the ID of the organization this endpoint belongs to.
```
woutslakhorst, line 252: rather see it move to service

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+* **type**: MUST be a URN describing the type of this endpoint in the form of `urn:oid:1.3.6.1.4.1.54851.2:type`.
+* **loc**: (location) MUST be a URL pointing to where the endpoint is exposed (e.g. `https://nuts.nl/some/path`)
+* **nbf**: (not before) MUST contain an RFC 3339 timestamp indicating from what date/time the endpoint SHOULD be considered valid.
+
+The following fields MAY be present:
+* **exp**: (expiration) MUST contain an RFC 3339 timestamp indicating when the endpoint expires and SHOULD NOT be considered valid anymore.
+* **svc**: (services) MUST contain an array with service IDs that this endpoint is linked to.
```
woutslakhorst, line 259: lol

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+### 4.4. Endpoint
+This document registers an endpoint for an organization which CAN be used to exchange data with that organization.
+
+#### Fields
+The following fields MUST be present:
+* **id**: MUST be a unique string identifying this endpoint.
+* **oid**: MUST contain the ID of the organization this endpoint belongs to.
+* **type**: MUST be a URN describing the type of this endpoint in the form of `urn:oid:1.3.6.1.4.1.54851.2:type`.
+* **loc**: (location) MUST be a URL pointing to where the endpoint is exposed (e.g. `https://nuts.nl/some/path`)
+* **nbf**: (not before) MUST contain an RFC 3339 timestamp indicating from what date/time the endpoint SHOULD be considered valid.
```
woutslakhorst, line 255: I don't know if this is needed when referencing from a service. A service will need it though..... A delete action is therefore also needed for endpoint/service removal

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+    "nbf": "2020-04-23T18:25:43.511Z",
+    "svc": ["397e9d20-7c1a-446f-a728-db3ed2452472"]
+}
+```
+
+#### Document type
+The document type when indicating an endpoint MUST be `registry/endpoint`.
+
+#### Signer
+This document MUST be signed by the organization's vendor.
```
woutslakhorst, line 277: rfc008 ref?

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+
+## 5. Vendor proof
```
woutslakhorst, line 279: what about credential types?

woutslakhorst (commented):
The proofs also closely relate to the signed contracts of RFC002. If there's overlap we might be able to construct a single _proof_ RFC...

rfc/rfc006-distributed-registry.md
reinkrul: After discussion: provide out-of-band denylist/crl

rfc/rfc006-distributed-registry.md
woutslakhorst: This is actually not needed, maybe in the logical layer, not in the lower layers. It'll be orphaned, it may be deleted from a relational model. If it would be referenced in the future, it can be via document reference.

rfc/rfc006-distributed-registry.md
reinkrul: After discussion: endpoints are floating (made by a vendor), services created under a organization have references to endpoints.

TODO: Add domain model to this RFC

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,344 @@
...
+
+### 4.1. Vendor
+This document registers a vendor on the registry. Since new vendors can't connect to other nodes in the network since their
+CA certificate isn't trusted yet, another vendor SHOULD register the new vendor through the following process:
+
+1. Vendor *A* sends vendor *B*:
+   1. the last document hash to use as *prev* and,
+   2. its vendor CA certificate so *B* will be able to connect to *A*.
+2. Vendor *B* generates a vendor CA certificate and sends it to *A*.
+3. Vendor *A* creates the vendor registrations with *A*'s CA certificate, then signs and publishes it.
```
reinkrul, line 89: with *A*'s CA certificate -> B's

reinkrul:
> The proofs also closely relate to the signed contracts of RFC002. If there's overlap we might be able to construct a single _proof_ RFC...

I'll spec roughly what I need for this RFC and then we can mix&match that with RFC002 and decide whether we need to refactor it to something shared (we probably need to).

rfc/rfc006-distributed-registry.md
reinkrul: Section removed

rfc/rfc006-distributed-registry.md
reinkrul: ?

rfc/rfc006-distributed-registry.md
reinkrul: Should probably be in RFC008

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,328 @@
...
+When a vendor loses or leaks its private key, the vendor SHOULD remove the certificate from its registration to avoid
+malicious parties exploiting it. However, other parties MUST regard signatures created before removal of the certificate
+ as valid. How the vendor SHOULD change its CA certificate depends on the situation:
+
+##### CA certificate private key loss
+In case the vendor loses its CA private key it MUST register a new vendor CA certificate. This can be done in two ways,
+depending on what was lost:
+
+1. If the vendor still has a valid signing certificate it COULD sign the update using that private key.
+2. Otherwise, when the signing certificate has expired or the private key is lost as well, the vendor MUST also provide
```
woutslakhorst, line 150: And publish via another Vendor? Wouldn't it be easier to just add a new _registry/vendor_ doc?

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,328 @@
...
+  The vendor ID SHALL NOT be changed; relation to a care organization SHALL NOT be transfered to another vendor. 
+* **name**: MUST contain the organization's commonly known name as string. Only alphanumeric, dashes (`-`) and space characters are allowed,
+  it SHALL NOT be empty or only contain spaces. It MUST conform to the following regex: `[a-zA-Z0-9 ]+`.
+* **svcs** (services): MUST contain an array of services this organization offers (see "Services" below).
+* **prfs** (proofs): MUST contain an array with cryptographic proofs asserting the organization as being an actual care organization.
+  at the same it serves as commitment from the care organization that it uses the vendor's (identified by **vid**) software.
+  See section 7 for supported proof types.
+
+##### Services
+A service groups a set of endpoints to be used for a specific use case (Bolt) for a specific organization. A service
```
woutslakhorst, line 200: Add Bolt to terminology

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,328 @@
...
+### 6.2. Agreements
+Networks COULD decide they want every vendor to sign an agreement when registering. In that case the vendor SHOULD specify
+the SHA-1 hash(es) of the agreement(s) that it commits to in the proof. The agreements SHOULD be published out of band
+on a well-known location. Other vendors SHOULD verify that the newly registered vendor specified an agreement they accept
+(accepted agreements COULD be configured in the vendor software).
+
+### 6.3. Accepted certificates
+This describes which certificates are accepted as vendor proof.
+
+#### 6.3.4. PKIoverheid Persoonlijk Organisatiegebonden Certificaat
```
woutslakhorst, line 294: numbering

rfc/rfc006-distributed-registry.md
reinkrul: True, since the entries can just be merged (as mentioned in 'Duplicates')

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,328 @@
...
+       { "op": "replace", "path": "/name", "value": "NoPlaceLikeHome Thuiszorg" }
+    ]
+}
+```
+
+When using a destructive operation (especially when using `remove` or `replace` on arrays) vendors COULD use the
+`test` operation to make sure the intended values are removed.
+
+## 4.2. Signing
+All documents MUST be signed by the vendor who owns the object (be it a vendor, organization or endpoint). The only
```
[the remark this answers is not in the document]
reinkrul: "Ownership" is a concept; when a party signs and publishes a document, that party owns it. Thus if an organization or endpoint is published, it is owned (and can only be updated) by the publishing party. If a vendor adds a new CA certificate, ownership is 'transferred' or 'shared' with that new CA certificate and it (or rather; signing certificates issued under that CA certificate) can be used to update the document.

Now I think of it, we register the CA certificate which can be rotated, so we need a concept (but is probably internal to the software) of 'entity' or 'party' (the vendor) which is authenticated by one or more CA certificates. Then an entity 'owns' a set of documents rather than a specific CA certificate. When a new CA certificate is registered it's added to the cert. set of that entity. Ownership can be transferred by simply adding the new certificates and removing the old.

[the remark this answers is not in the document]
woutslakhorst: That's true

rfc/rfc006-distributed-registry.md
woutslakhorst: True, the ID should come from a third party (from certificate) which would it make to do an ID comparison?

rfc/rfc006-distributed-registry.md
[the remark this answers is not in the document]
reinkrul: Decision: when a vendor re-registers itself (because it has lost its CA and signing keys) it must do so under the same ID (with a new vendor CA certificate and proof). From that point in time (or rather, in DAG) the new CA certificate is added to the pool of CA certificates that authenticate the vendor, meaning it can update documents that were originally published under the old CA certificate.

rfc/rfc006-distributed-registry.md
reinkrul: Added in spec. Note that **PKIOverheid Organisatie**-certificates don't contain a KVK-nummer (according to the KPN Helpdesk) so we still have a challenge to solve there...

rfc/rfc006-distributed-registry.md
```diff
@@ -316,7 +319,7 @@ The JWT SHALL NOT be encrypted.
 ### 7.1. Accepted certificates
```
woutslakhorst, line 319: pkio would be acceptable here as well

woutslakhorst (commented):
A reference on how RFC001 §4.4 is implemented would be a welcome addition

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,331 @@
...
+* **vid** (vendor ID): MUST be the organization's software vendor's ID as URN (e.g. `urn:oid:1.3.6.1.4.1.54851.4:1234`).
+  The vendor ID SHALL NOT be changed; relation to a care organization SHALL NOT be transfered to another vendor. 
+* **name**: MUST contain the organization's commonly known name as string. Only alphanumeric, dashes (`-`) and space characters are allowed,
+  it SHALL NOT be empty or only contain spaces. It MUST conform to the following regex: `[a-zA-Z0-9 ]+`.
+* **svcs** (services): MUST contain an array of services this organization offers (see "Services" below).
+* **prfs** (proofs): MUST contain an array with cryptographic proofs asserting the organization as being an actual care organization.
+  at the same it serves as commitment from the care organization that it uses the vendor's (identified by **vid**) software.
+  See section 7 for supported proof types.
+
+##### Services
```
woutslakhorst, line 202: Maybe services need to be a different object type. A service can be turned on and off per organization. This would be done independently from the organization registration. A question that also arises: must a organization sign something when a service is enabled? In the current workflow, the vendor handles this for the organization.

Services also require a nbf

reinkrul: After discussion; Service will be modeled as separate object. Since we currently don't have practical authentication means to use as organization proof, proofs will be added in a later version.

rfc/rfc006-distributed-registry.md
```diff
@@ -241,26 +226,64 @@ It's identified by the type `registry/endpoint`.
...
 ```
 
+### 5.4. Service
+A service groups a set of endpoints to be used for a specific use case (Bolt) for a specific organization. It's identified by the type `registry/service`.
+It MUST contain the following fields:
+* **oid** (organization ID): MUST contain the ID of the organization this service belongs to.
+* **vid** (vendor ID): MUST contain the ID of the vendor this endpoint belongs to.
+* **name**: MUST contain the service's case sensitive name. It MUST be unique within the organization and vendor combination.
+  It is used by other parties who want to execute a specific Bolt (with this organization) to look up the correct endpoints.
+* **eps** (endpoints): MUST contain an array of endpoint IDs to be used for this service. The endpoints MUST belong to
```
woutslakhorst, line 257: or just have the same `vid`

rfc/rfc006-distributed-registry.md
```diff
@@ -241,26 +226,64 @@ It's identified by the type `registry/endpoint`.
...
+  "oid": "urn:oid:2.16.840.1.113883.2.4.6.1:06123456",
+  "vid": "urn:oid:1.3.6.1.4.1.54851.4:1234",
+  "name": "some-bolt",
+  "eps": ["1B5E2A91-5C19-41B6-9709-BC9050D193E5", "some-other-endpoint-ID"],
+  "nbf": "2020-10-20T20:30:50.52Z"
+}
+```
+
+#### Validation
+When processing an organization registration or update it MUST be validated as follows:
```
woutslakhorst, line 276: organization == service

rfc/rfc006-distributed-registry.md
```diff
@@ -241,26 +226,64 @@ It's identified by the type `registry/endpoint`.
...
+
+1. Assert all required fields are present.
+2. Assert all fields value formats are valid.
+3. For updates: assert only modifiable fields are updated (only **eps** is modifiable).
```
woutslakhorst, line 280: isn't `exp` modifiable?

woutslakhorst (changes requested):
What about using VerifiableProofs for the proofs?

rfc/rfc006-distributed-registry.md
```diff
@@ -0,0 +1,378 @@
...
+* **loc** (location): MUST be a URL pointing to where the endpoint is exposed (e.g. `https://nuts.nl/some/path`)
+* **nbf** (not before): MUST contain an RFC 3339 timestamp indicating from what date/time the endpoint SHOULD be considered valid.
+
+The following fields MAY be present:
+* **exp** (expiration): MUST contain an RFC 3339 timestamp indicating when the endpoint expires and SHOULD NOT be considered valid anymore.
+
+Example:
+```json
+{
+    "id": "de59b062-d783-4727-a05e-57ed6035f00d",
```
woutslakhorst, line 242: just plain ID or urn?

reinkrul: free-format ID since it must be unique just for the vendor

woutslakhorst (approved):

PR #11, opened 2020-12-01
RFC008 - Added EKU requirements to CA

woutslakhorst:
CA's must have the same EKU as client/server certs

### DIFF
--- a/rfc/rfc008-certificate-structure.md
+++ b/rfc/rfc008-certificate-structure.md
@@ -72,6 +72,7 @@ This CA signs TLS certificates and signing certificates. This is a very importan
 | Validity | 3 years |
 | Is CA? | true |
 | Key Usage | certSign |
+| Extended Key Usage | serverAuthentication, clientAuthentication |
 | SubjectAlternativeName | otherName=1.3.6.1.4.1.54851.4: |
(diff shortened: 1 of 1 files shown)


reinkrul:
I think I missed the discussion, why?

woutslakhorst:
Something I discovered: apparently if you have a certificate with an EKU, your CA also has to have it....
The go `cert.Verify` check could fail on it.

reinkrul:
> Something I discovered: apparently if you have a certificate with an EKU, your CA also has to have it....
> The go `cert.Verify` check could fail on it.

This is quite peculiar. RFC5280 states (https://tools.ietf.org/html/rfc5280#section-4.2.1.12):

> In general, this  extension will appear only in end entity certificates. 

However the CA Browser Forum states in their baseline requirements (https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.7.3.pdf):

> 7.1.5 Name constraints
For a Subordinate CA Certificate to be considered Technically Constrained, the certificate
MUST include an Extended Key Usage (EKU) extension specifying all extended key usages
that the Subordinate CA Certificate is authorized to issue certificates for. The
anyExtendedKeyUsage KeyPurposeId MUST NOT appear within this extension.
If the Subordinate CA Certificate includes the id-kp-serverAuth extended key usage, then
the Subordinate CA Certificate MUST include the Name Constraints X.509v3 extension with
constraints on dNSName, iPAddress and DirectoryName as follows:

So the root CA *may*  specify extKeyUsage, but if we have a sub-CA is MUST include NameConstraints. Though it doesn't say anything about non-subordinate-CAs.

> The go cert.Verify check could fail on it.

_Could_ fail or always fails?

reinkrul (approved):

PR #23, opened 2021-01-13
RFC010 multi vendor

### LINKED ISSUE #18: New RFC: how to cater care organizations with multiple vendors, addition to RFC006

woutslakhorst:
RFC010 defines how a DID subject can authenticate with another DID subject when the DID subjects are, in fact, the same organization.

Closes #18


rfc/rfc010-multiple-operators.md
```diff
@@ -0,0 +1,64 @@
...
+|  |  |
+| :--- | :--- |
+| Nuts foundation | W.M. Slakhorst |
+| Request for Comments: 010 | Nedap |
+|  | January 2021 |
+
+## Multiple operator service access
+### Abstract
+
+This RFC describes an addition to [RFC003](rfc003-oauth2-auithorization.md) and [RFC006](rfc006-distributed-registry.md) on how to handle resource access when a legal party has multiple node operators. This is the case when a care organization uses multiple pieces of software from different SaaS providers. A SaaS provider can alter the care organization DID document so the DID document of the other provider is added as synonym. This is done by using the *also-known-as* relation.
```
reinkrul, line 12: maybe define SaaS a little bit in the context? Something like

` SaaS provider (party operating a Nuts node as a service on behalf of their clients (care organizations)`

reinkrul, line 12: > This is the case when a care organization uses multiple pieces of software from different SaaS providers

'multiple pieces of software' can go here I think, just 'uses multiple SaaS providers'? Since they could be using different implementations of the Nuts spec, or they don't, it doesn't matter. What matters is that they're operated by different companies.

rfc/rfc010-multiple-operators.md
```diff
@@ -0,0 +1,64 @@
+# RFC010 Multiple operator service access
```
reinkrul, line 1: I don't like the term 'piece of software' because it sounds like 'spreektaal'. Maybe 'software system' or just 'system'?

rfc/rfc010-multiple-operators.md
```diff
@@ -0,0 +1,64 @@
...
+This document is currently a draft.
+
+### Copyright Notice
+![](../.gitbook/assets/license.png)
+
+This document is released under the [Attribution-ShareAlike 4.0 International \(CC BY-SA 4.0\) license](https://creativecommons.org/licenses/by-sa/4.0/).
+
+## 1. Introduction
+
+Care organizations rarely use one piece of software. Larger organizations can use up to a hundred different pieces of software. Every software vendor that has signed a **Data Processing Agreement** with the care organization is allowed to transfer data to any other software vendor within the limits of the **Data Processing Agreement**.
```
reinkrul, line 25: " Larger organizations can use up to a hundred different pieces of software" and might be using more than 100, so I'd just leave this sentence away.

rfc/rfc010-multiple-operators.md
```diff
@@ -0,0 +1,64 @@
...
+
+Care organizations rarely use one piece of software. Larger organizations can use up to a hundred different pieces of software. Every software vendor that has signed a **Data Processing Agreement** with the care organization is allowed to transfer data to any other software vendor within the limits of the **Data Processing Agreement**. 
+
+To support this functionality, some registration and extended validation is required throughout the process of data transfer.
+
+## 2. Terminology
+
+* **Bolt**: a use case built on top of the functionality Nuts provides.
+* **DID**: a decentralized identifier according to the [core-did-spec](https://www.w3.org/TR/did-core). 
+* **DPA**: Data Processing Agreement as defined by the GDPR.
```
reinkrul, line 33: GDPR is not defined here

rfc/rfc010-multiple-operators.md
```diff
@@ -0,0 +1,64 @@
...
+
+Other terminology is taken from the [Nuts Start Architecture](rfc001-nuts-start-architecture.md#nuts-start-architecture).
+
+## 3. Registration
+
+Each care organization that is registered at a node by a SaaS provider will have its own DID document. If a care organization uses 2 SaaS providers, it'll have two different DID documents. Since identification within the OAuth flow is done by DID, the DID subject need to be aware of each other. This can be done by adding the DID of one document and add it to the other. This is done by using the `alsoKnownAs` field defined in the [core-did-spec](https://www.w3.org/TR/did-core/#also-known-as).
```
reinkrul, line 39: DID subject -> subjects

reinkrul, line 39: explain what is identified here, something like;

> Since identification of the care organization within the OAuth flow

reinkrul, line 39: > This can be done by adding the DID of one document and add it to the other

"adding the DID of one document" already implies implementation, so I'd use "referring" in this sentence, since the implementation is defined in the next sentence (using alsoKnownAs)

This can be done by having the DID documents of the SaaS providers refer to each other.

rfc/rfc010-multiple-operators.md
```diff
@@ -0,0 +1,64 @@
...
+{
+  "@context": [ "https://www.w3.org/ns/did/v1" ],
+  "id": "did:nuts:1",
+  "alsoKnownAs": [
+    "did:nuts:2"
+  ]
+}
+```
+
+This is a unary relation. In the above example, `did:nuts:2` does not know anything about `did:nuts:1`.
```
reinkrul, line 53: an unary

rfc/rfc010-multiple-operators.md
```diff
@@ -0,0 +1,64 @@
...
+The controller of the DID document MUST verify that it added the correct DID in the `alsoKnownAs` field.
+
+## 4. Authentication
+
+As defined in point 3 in [§6.2 of RFC003](rfc003-oauth2-auithorization.md#6-2-authorization), if the actor and custodian are the same, access is granted. If a custodian had added an actor in their DID document under the `alsoKnownAs` field, it MUST grant access as defined by point 3 in §6.2 of RFC003. Bolts MUST describe what resources may be accessed in this way and if user context is required.
```
reinkrul, line 58: If a custodian has added

> "Bolts MUST describe what resources may be accessed in this way and if user context is required"
So all Bolts MUST define it, or do we have default behaviour (since the current Bolts obviously don't spec it)

rfc/rfc010-multiple-operators.md
```diff
@@ -0,0 +1,64 @@
...
+
+## 5. Hijacked DID documents 
+
+A DID subject that is added under the `alsoKnownAs` field is granted access to resources from the custodian.
+In the case when the private key of the actor is compromised and the attacker is changing keys, the custodian SHOULD take appropriate action.
+The node operator MUST remove a DID from the `alsoKnownAs` field when the `authentication` or `controller` field changes in the referenced DID document.
```
reinkrul, line 64: > The node operator MUST remove a DID from the `alsoKnownAs` field when the `authentication` or `controller` field changes in the referenced DID document.

I find it a bit hard to read without a concrete example (say vendor Alice and Bob or A and B), but if I understand correctly if A refers to B's DID using `alsoKnownAs`, A MUST remove the `alsoKnownAs` referral to B when B changes its authentication or controller key, regardless there was an attack or not?

reinkrul (changes requested):

rfc/rfc010-multiple-operators.md
stevenvegt, line 12: Perhaps elaborate the problem and solution here a bit more explicit. Talk about te trust relation between node provider and care provider. Than explicitly introduce the solution this RFC will be about: 
>This RFC introduces a solution where DID documents refer to each other using the also-known-as propertie.

rfc/rfc010-multiple-operators.md
stevenvegt: Perhaps explain why each saas provider has their own DID document? I suspect it has to do with the coupling between document and services which are vendor specific?

rfc/rfc010-multiple-operators.md
stevenvegt, line 53: So, why do we nog require the other document to point back?

rfc/rfc010-multiple-operators.md
stevenvegt, line 58: The other document is just a number until it has VCs. Why not check fot those? This also prevents typos were the wrong documents are coupled and everything is mixed up.

rfc/rfc010-multiple-operators.md
stevenvegt, line 64: This all feels very error prone. I would like to hook in to existing processes like revoked VCs for example.

rfc/rfc010-multiple-operators.md
stevenvegt, line 1: I thing this document describes a specific subset where the care provider does not administers its own DID document. I think we should add that case here as well or make it explicit that it is about delegated identity.

stevenvegt (changes requested):

rfc/rfc010-multiple-operators.md
woutslakhorst: added it to terminology

rfc/rfc010-multiple-operators.md
woutslakhorst: rewrote everywhere (1 instance :p)

rfc/rfc010-multiple-operators.md
woutslakhorst: yes examples are not needed in specs.

rfc/rfc010-multiple-operators.md
woutslakhorst: Added term and link.

rfc/rfc010-multiple-operators.md
woutslakhorst: added a bit more explanation.

woutslakhorst: Switched to subject and added subject to terminology

rfc/rfc010-multiple-operators.md
woutslakhorst: nope, words like `unary` and `user` start with a Y pronunciation and thus it's **a**

rfc/rfc010-multiple-operators.md
woutslakhorst: added: If nothing is described => no access

rfc/rfc010-multiple-operators.md
woutslakhorst: That's correct and it's quite harsh. I'm not sure if it's needed. We should discuss it.

rfc/rfc010-multiple-operators.md
woutslakhorst: Why would we? Relations are unary in data transfer between custodian and actor. The custodian needs to trust the actor.

The other way around could be needed to signal the actor that there's data available on another node. That could be a recommendation not a necessity.

rfc/rfc010-multiple-operators.md
woutslakhorst: This spec only defines the underlying mechanism for authz/trust. Specifications of VC's could define the automatic additions of synonyms when data of the VCs match.

rfc/rfc010-multiple-operators.md
woutslakhorst: This spec has nothing to do with VCs. This point was added to detect compromised DID Documents and protect the custodian automatically. Automatic synonyms via VCs could potentially mitigate the problem of user intervention.

woutslakhorst: Will change `node operator` to `node`, since it's the software that should do it.

rfc/rfc010-multiple-operators.md
woutslakhorst: It's still valid for when an care organization manages its own keys. They will probably not use the functionality. The only thing this spec defines is that data MUST be accessible for DID subjects that refer to each other using the `also-known-as` property.

woutslakhorst:
This mechanism poses a security risk through fishing. Giving access to data must be supported with a piece of crypto (VCs). The idea for creating synonyms is OK, but it must be done in a different way.

PR #34, opened 2021-03-02
RFC006: Specified how Nuts Network endpoints are published as DID services

reinkrul:


rfc/rfc006-distributed-registry.md
```diff
@@ -333,6 +333,13 @@ The care organisation refers to it:
 }
 ```
 
+### 4.1. Network Endpoint Service
+
+The underlying peer-to-peer network does not automatically discover new peers but it relies on the system to tell
+which peers to connect to instead. Nodes SHOULD publish their Nuts Network endpoint(s) as a service in their DID Document(s).
+The `type` of this service MUST be `nuts-network-grpc`. The `serviceEndpoint` MUST be a single gRPC endpoint other
+nodes can use to connect to the node. In MUST be in the format of `<host>:<port>`. A node MAY publish multiple network endpoints.
```
stevenvegt, line 341: typo: In -> It

stevenvegt, line 341: Since the service datastructer is very generic, I think it helpfull to add an example service with one or multiple endpoints, or describe the format better as an array of maps?

stevenvegt (changes requested):

rfc/rfc006-distributed-registry.md
reinkrul: As you say, the datastructure is very generic, that's why I spec'd it as very simple and strict: MUST be a single gRPC endpoint. But I can add an example though.

reinkrul: I see I don't honor the compound service spec as stated above in `4.` so I'll add that.

woutslakhorst:
This PR is part of a larger discussion of the identity of the node vs DID

rfc/rfc006-distributed-registry.md
```diff
@@ -333,6 +333,25 @@ The care organisation refers to it:
...
+nodes can use to connect to the node. It MUST be in the format of `<host>:<port>`. A node MAY publish multiple network endpoints
+through multiple services with the `NutsNetworkGRPCService` type.
+
+Example:
+
+```json
+{
+  "id": "did:nuts:abc#NutsNetworkGRPCService-1",
+  "type": "NutsNetworkGRPCService",
+  "serviceEndpoint": "nuts.carevendor.nl:5555"
```
stevenvegt, line 350: Better change this to an official example domain.

rfc/rfc006-distributed-registry.md
```diff
@@ -333,6 +333,25 @@ The care organisation refers to it:
 }
 ```
 
+### 4.1. Network Endpoint Service
+
+The underlying peer-to-peer network does not automatically discover new peers, but it relies on the system to tell
+to which peers to connect. Nodes SHOULD publish their Nuts Network endpoint(s) as a service in their DID Document(s).
```
stevenvegt, line 339: Do you mean EVERY DID Document made by a certain node should contain this endpoint? Or some? And which ones?

stevenvegt (changes requested):

rfc/rfc006-distributed-registry.md
woutslakhorst: Important question: does this require a relation between the certificate and the DID publishing it? If not: then can any DID doc contain such an endpoint regardless of the DID is managed by that node?

woutslakhorst:
not now! Will be solved at a later moment.

PR #36, opened 2021-03-10
replace timeline concept with generic updates&ordering

### LINKED ISSUE #30: RFC004: remove timeline ID and version
was required for patch style updates

woutslakhorst:
Closes #30

The choice for parallel updates is determined by information of the transaction compared to application data and previous transactions.

The proposed changes should be checked with application logic currently in the master branch of the node.

### DIFF
--- a/rfc/rfc004-verifiable-transactional-graph.md
+++ b/rfc/rfc004-verifiable-transactional-graph.md
@@ -110,8 +110,9 @@ Consider the DAG from the previous chapter. If transaction `C` and  `D` where to
 This would create an inconsistency in the network. To fix this the following rules MUST be taken into account:
 
 * If the payloads are equal, mark both transactions as processed and apply one of the payloads.
-* If the payloads are not equal, apply the payload with the highest *iat*.
-* Else apply the payload with the highest hash.
+* If the payloads are not equal, merge payload according to rules for the specific payload.
+
+The last rule requires the payload to be immutable or conform to its contents MUST be composed of [conflict-free replicated data types](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type)  
 
 ### 3.5. Processing the DAG
 
--- a/rfc/rfc004-verifiable-transactional-graph.md
+++ b/rfc/rfc004-verifiable-transactional-graph.md
@@ -110,9 +110,9 @@ Consider the DAG from the previous chapter. If transaction `C` and  `D` where to
 This would create an inconsistency in the network. To fix this the following rules MUST be taken into account:
 
 * If the payloads are equal, mark both transactions as processed and apply one of the payloads.
-* If the payloads are not equal, merge payload according to rules for the specific payload.
+* If the payloads are not equal, create a representation that is a merger of the payloads according to rules for the specific payload.
 
-The last rule requires the payload to be immutable or conform to its contents MUST be composed of [conflict-free replicated data types](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type)  
+The last rule requires the payload to be immutable or conform to its contents MUST be composed of [conflict-free replicated data types](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type)
 
 ### 3.5. Processing the DAG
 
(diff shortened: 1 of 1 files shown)


stevenvegt (approved):

rfc/rfc004-verifiable-transactional-graph.md
```diff
@@ -107,25 +100,18 @@ The following orders are invalid:
...
-* Assert that the update has been issued by the owner \(signer\) of the original transaction.
-* Gaps in timeline versioning MAY indicate the consumer is missing a \(branching\) transaction.
-* Duplicates in timeline versioning MAY indicate a race condition in which the producer updated the application data based on out-of-date state.
-  * When the payloads are equal one of the updates COULD be applied and the other one ignored.
-  * When the payloads differ the transaction with the lowest **iat** SHOULD be applied first.
+Consider the DAG from the previous chapter. If transaction `C` and  `D` where to update the same application data, nodes could process the transaction in a different order.
+This would create an inconsistency in the network. To fix this the following rules MUST be taken into account:
 
-    If **iat**s are equal the transaction with the lowest hash should be applied first.
+* If the payloads are equal, mark both transactions as processed and apply one of the payloads.
```
reinkrul, line 112: This change means we have to (severely) revise the DAG walking algorithm.

Also a complication: since the RFC004 layer now doesn't know what transactions are about the same "application data" (previously it did through `timeline ID`), the algorithm can't determine what transactions should be taken into consideration to determine the order.

Another problem which exists in the current implementation as well but doesn't get fixed with this change is that the transaction ordering (of already applied transactions)  might change:
When a new transaction with a older `iat` gets published as branch of an older transaction, the transactions might have to be replayed. This could be a problem when DID `authentication` keys are removed, which might invalidate future versions of that DID document which were previously considered valid.

Not sure what the solution is without making DAG traversal when history changes massively complicated.

We might push that responsibility into the application layer: VDR could "insert" that version where it belongs (according to `iat`) and validate all succeeding versions from that point on? Complication is that information derived from DIDs (VCs) could be impacted/invalidated too, and should be reprocessed as well, further complicating things.

On the other hand, since every update is a full replacement it might not matter that much.

woutslakhorst: I see the point. VCs might have been issued with a key that is no longer valid a few seconds later when a new DID Document update has been processed. But don't you have the same problem when two transactions use the same timeline version number. According to the current rules a similar sorting applies?

woutslakhorst: And it's only a problem for docs with the exact same `iat`

woutslakhorst:
The issue has a more in-depth discussion on the problem.

rfc/rfc004-verifiable-transactional-graph.md
woutslakhorst: @reinkrul I changed the line. So we don't have to change the DAG algorithm (yet). All other remarks are unrelated to this PR?

woutslakhorst:
@reinkrul re-evaluate?

reinkrul (approved):

rfc/rfc004-verifiable-transactional-graph.md
```diff
@@ -107,25 +100,19 @@ The following orders are invalid:
...
-* Assert that the update has been issued by the owner \(signer\) of the original transaction.
-* Gaps in timeline versioning MAY indicate the consumer is missing a \(branching\) transaction.
-* Duplicates in timeline versioning MAY indicate a race condition in which the producer updated the application data based on out-of-date state.
-  * When the payloads are equal one of the updates COULD be applied and the other one ignored.
-  * When the payloads differ the transaction with the lowest **iat** SHOULD be applied first.
+* If the payloads are equal, process as normal.
+* If the payloads are not equal, create a representation that is a merger of the payloads according to rules for the specific payload.
 
-    If **iat**s are equal the transaction with the lowest hash should be applied first.
+The last rule requires the payload to be immutable or conform to its contents MUST be composed of [conflict-free replicated data types](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type)
```
woutslakhorst, line 115: grammar

PR #40, opened 2021-03-12
RFC006: Protection to public key zerging

reinkrul:
Challenge: Can we refine it so we have protection against many keys in many updated documents, too?
Solution: fix transaction flooding in RFC005

### DIFF
--- a/rfc/rfc006-distributed-registry.md
+++ b/rfc/rfc006-distributed-registry.md
@@ -285,8 +285,6 @@ By registering a very large number of public keys in a DID Document an attacker 
 
 To prevent this from happening a DID Document MUST contain no more than 50 keys.
 
-TODO: What if the attacker just creates 100.000 versions of a document with in each 50 new keys?
-
 #### Threat: Resource Exhaustion through Very large number of keys within DID (max. 50 keys)
 
 ### 3.4 Privacy considerations
--- a/rfc/rfc006-distributed-registry.md
+++ b/rfc/rfc006-distributed-registry.md
@@ -255,7 +255,7 @@ Almost all security considerations are covered by the mechanisms described in [R
 - **deletion** - All DID documents are published and copied on a mesh network. Deletion of a single document will only occur locally and will not damage other nodes.
 - **modification** - DID documents can only be modified if they are published with a signature from one of the `authentication` keys.
 - **man-in-the-middle** - All communications is sent over two-way TLS and all documents are signed. A DID can not be hijacked since it is derived from the public key.
-- **denial of service** - This is out of scope and handled by [RFC004](rfc004-verifiable-transactional-graph.md).
+- **denial of service** - This is mostly out of scope and handled by [RFC004](rfc004-verifiable-transactional-graph.md), unless it can only be handled by this RFC.
 
 #### 3.3.1 Protection against DID hijacking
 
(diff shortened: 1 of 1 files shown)


rfc/rfc006-distributed-registry.md
```diff
@@ -279,6 +279,16 @@ When control over a DID document has been lost, the DID subject will have to hav
 Without Verifiable Credentials linked to the DID document, the DID document no longer has any value.
 The DID subject will have to go through the process of reacquiring all Verifiable Credentials for a new DID document.
 
+#### 3.3.4 Protection against public key zerging
+
+By registering a very large number of public keys in a DID Document an attacker can exhaust resources used for storing public keys.
+
+To prevent this from happening a DID Document MUST contain no more than 50 keys.
+
+TODO: What if the attacker just creates 100.000 versions of a document with in each 50 new keys?
```
woutslakhorst, line 288: That would require 100.00 transactions which are noticeable and can be prevented.

reinkrul: Agreed, that should be an identified threat in RFC004 (transaction flooding)

rfc/rfc006-distributed-registry.md
```diff
@@ -279,6 +279,14 @@ When control over a DID document has been lost, the DID subject will have to hav
 Without Verifiable Credentials linked to the DID document, the DID document no longer has any value.
 The DID subject will have to go through the process of reacquiring all Verifiable Credentials for a new DID document.
 
+#### 3.3.4 Protection against public key zerging
+
+By registering a very large number of public keys in a DID Document an attacker can exhaust resources used for storing public keys.
+
+To prevent this from happening a DID Document MUST contain no more than 50 keys.
+
+#### Threat: Resource Exhaustion through Very large number of keys within DID (max. 50 keys)
```
woutslakhorst, line 288: why this header?

rfc/rfc006-distributed-registry.md
```diff
@@ -279,6 +279,14 @@ When control over a DID document has been lost, the DID subject will have to hav
 Without Verifiable Credentials linked to the DID document, the DID document no longer has any value.
 The DID subject will have to go through the process of reacquiring all Verifiable Credentials for a new DID document.
 
+#### 3.3.4 Protection against public key zerging
```
stevenvegt, line 282: Is `key zerging` a thing? Could not find anything outside of game context.

rfc/rfc006-distributed-registry.md
```diff
@@ -279,6 +279,14 @@ When control over a DID document has been lost, the DID subject will have to hav
...
+
+By registering a very large number of public keys in a DID Document an attacker can exhaust resources used for storing public keys.
+
+To prevent this from happening a DID Document MUST contain no more than 50 keys.
```
stevenvegt, line 286: 50 feels a bit arbitrary. Perhaps limit the size of the network payload instead of the amount of keys?

stevenvegt (changes requested):

rfc/rfc006-distributed-registry.md
woutslakhorst: it has already made the urban dictionary: _Using massive amounts of weak, or generic units, to attack an enemy._

reinkrul: I'm not sure whether a term has made it to Urban Dictionary is a good measure :'-D

rfc/rfc006-distributed-registry.md
woutslakhorst: that would also be arbitrary.

A cluster of 5 nodes with their own keys would yield 5 keys. When rotating keys it would be doubled: 10.
So 10 is a well though minimum.
50 is a multiple of that and yes it's arbitrary.

woutslakhorst:
Will be continued in #49

PR #58, opened 2021-04-29
changed to DID generation

woutslakhorst:
This makes it way easier to compare keys with DIDs.

Also RFC7638 is easier to understand when using different keys.

### DIFF
--- a/rfc/rfc006-distributed-registry.md
+++ b/rfc/rfc006-distributed-registry.md
@@ -72,18 +72,18 @@ For example, consider the following Ed25519 key (as JWK):
 
 ```json
 {
-  "kty" : "OKP",
-  "crv" : "Ed25519",
-  "x"   : "11qYAYKxCrfVS_7TyWQHOg7hcvPapiMlrwIaaPcHURo",
-  "use" : "sig",
-  "kid" : "FdFYFzERwC2uCBB46pZQi4GG85LujR8obt-KWRBICVQ"
+  "kty" : "EC",
+  "crv" : "P-256",
+  "x"   : "Qn6xbZtOYFoLO2qMEAczcau9uGGWwa1bT+7JmAVLtg4=",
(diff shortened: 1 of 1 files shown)


rfc/rfc006-distributed-registry.md
```diff
@@ -64,7 +64,9 @@ base58char = "1" / "2" / "3" / "4" / "5" / "6" / "7" / "8" / "9" / "A" / "B" / "
 
 The `idstring` is derived from the public part of a key pair that was used to sign the transaction that created the DID document.
 
-`idstring = BASE-58(SHA-256(raw-public-key-bytes))`
+`idstring = BASE-58(hash)`
+
+The `hash` is calculated conforming [rfc7638](https://tools.ietf.org/html/rfc7638), using the `SHA256` hashing algorithm.
```
reinkrul, line 69: conform to, drop the comma

Do we need to specify the base58 alphabet?

rfc/rfc006-distributed-registry.md
```diff
@@ -64,7 +64,9 @@ base58char = "1" / "2" / "3" / "4" / "5" / "6" / "7" / "8" / "9" / "A" / "B" / "
...
 
 For example, consider the following Ed25519 key (as JWK):
```
reinkrul, line 72: maybe we should leave out the `kid` in this example?

reinkrul (changes requested):

rfc/rfc006-distributed-registry.md
woutslakhorst: it already is at line 57....

rfc/rfc006-distributed-registry.md
woutslakhorst: Its just to show that not all props are used in the thumbprint

reinkrul (approved):

PR #61, opened 2021-04-29
Regenerated spec relations image

reinkrul:


woutslakhorst:
maybe a screenshot so we can actually see what has changed?

reinkrul:
> maybe a screenshot so we can actually see what has changed?

When you switch to "rich diff" it renders "before" and "after" :-)

woutslakhorst (approved):
learned something 👍

PR #75, opened 2021-05-25
RFC005: Specified diagnostical request/response message

### LINKED ISSUE #69: RFC005: exchange nodes' diagnostic information
When the reference implementation's diagnostic information indicates a peer's DAG isn't up-to-date it's very hard (or impossible) to find out why. Allowing to find out more about the specific peer's software would help analysing these situations. This could be done by defining a query for a peer's diagnostic information.

reinkrul:
Fixes #69

### DIFF
--- a/.gitbook/assets/rfc005/network.proto
+++ b/.gitbook/assets/rfc005/network.proto
@@ -38,6 +38,8 @@ message NetworkMessage {
         TransactionList TransactionList = 102;
         TransactionPayloadQuery transactionPayloadQuery = 103;
         TransactionPayload transactionPayload = 104;
+        DiagnosticsRequest diagnosticsRequest = 105;
+        DiagnosticsResponse diagnosticsResponse = 106;
     }
 }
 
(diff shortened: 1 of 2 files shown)


woutslakhorst (changes requested):
what about `version`, `branch` and `commit`?

reinkrul:
> what about `version`, `branch` and `commit`?

Oh I missed "version". I intent to just add "version" and spec it as "either a (Git) commit ID or a semantic version", since it depends on the implementation.

I don't think we need `branch` and `commit` because when using Git (I'd think 99,99% of the Nuts nodes) the commit ID (contained in "version") will be the same regardless the branch. And you can determine the version given the commit ID when running the reference implementation. We might need to add something like `implementationName` (bad name though), then we can add a link to the GH repo (and a proprietary implementation can add their name, like `Slakhorst Node`)

woutslakhorst:
👍  indeed `name` + `semantic version` would be more generic and not implementation specific.

woutslakhorst (approved):

reinkrul:
Implementation issue: https://github.com/nuts-foundation/nuts-node/issues/308

PR #85, opened 2021-06-17
update DID Document deactivation

### LINKED ISSUE #37: RFC006: Update Revocation with services and other keys
Update [RFC006 section 3.2.4](Retrieve Unlocks) Delete, with how to handle services and other verificationMethods such as assertionMethods.

Proposal is to update the document with an empty document which only contains the DID id.
Explain that information still in the DID can never be changed and will become out of date, so delete everything. If you want to know more, look back to previous versions of the DID.

woutslakhorst:
Closes #37
next: nuts-foundation/nuts-node#341

### DIFF
--- a/rfc/rfc006-distributed-registry.md
+++ b/rfc/rfc006-distributed-registry.md
@@ -237,13 +237,8 @@ Example JWS header:
 
 #### 3.2.4 Delete (Deactivate)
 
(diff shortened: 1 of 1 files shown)


reinkrul (approved):

woutslakhorst:
deactivation through the DocManipulator already does what is required.

PR #87, opened 2021-06-18
RFC005: Specified duplicate connection handling

reinkrul:
Currently we've specified that duplicate connections must be avoided, but we haven't spec'd how. This lead to a bug in the Nuts Node.

This change specifies what to do when a second (duplicate) connection occurs.


woutslakhorst (commented):
For now, you could identify nodes by peerID + certificate. This prevents the falsified peerID attack.

reinkrul:
Discussed with @woutslakhorst @stevenvegt the essence of this PR is how we handle a duplicate connection (disconnect first one) from a peer. How we identify a peer and how we authenticate that identifier (peer ID) is out of scope and part of solving the "publisher identity" issue (https://github.com/nuts-foundation/nuts-specification/issues/47).

woutslakhorst (approved):

PR #99, opened 2021-06-30
RFC006: Expand service references

### LINKED ISSUE #78: RFC006: Reference services
Introducing references in services to other services.
When the URI of a service consists of a reference, this must be resolvable to another endpoint.
Then end of the chain should be either a DID or a URL.

A ref is a cobination of a _target DID_ document combined with a _type_: `ref:${target}/serviceEndpoint/type?=${type}`

`ref:did:nuts:123/serviceEndpoint?type=eOverdracht`

References can point to other references with a limited depth.
Implementors of a resolver should follow all these references and return the Endpoint.

woutslakhorst:
Closes #78 

URIs that start with `did` are references.
The limitation of a compound service that needs to reference absolute endpoints has been removed.

### DIFF
--- a/rfc/rfc006-distributed-registry.md
+++ b/rfc/rfc006-distributed-registry.md
@@ -276,8 +276,8 @@ A service can define an absolute endpoint URI or be a compound service.
 A service can also refer to other services.
 This is often the case when a SaaS provider defines endpoints to be used for all clients.
 
-For an absolute endpoint URI the `serviceEndpoint` MUST be a string containing the URI. 
-For a compound service the `serviceEndpoint` MUST contain a map with absolute endpoint URIs or references to other services. 
+For an absolute endpoint URI the `serviceEndpoint` MUST be a string containing a URL. 
+For a compound service the `serviceEndpoint` MUST contain a map with absolute endpoint URLs or references to other services. 
 References MUST be by query and not by fragment. Only `type` can be used as query param and it refers to the `type` field in a service. A compound service MAY refer to absolute endpoints from other DID Documents. See [§3.2 of did-core](https://www.w3.org/TR/did-core/#did-url-syntax) for DID URL syntax and [RFC3986](https://tools.ietf.org/html/rfc3986) for generic URL standards.
 
 A DID Document MAY NOT contain more than one service with the same type.
--- a/rfc/rfc006-distributed-registry.md
+++ b/rfc/rfc006-distributed-registry.md
@@ -367,9 +367,22 @@ An example of a deeply nested structure:
     {
       "id": "did:nuts:abc#4",
       "type": "oauth_prod",
-      "serviceEndpoint": "https:\\auth.example.com"
+      "serviceEndpoint": "https://auth.example.com"
     }
   ]
+}
+```
+
+Resolving `did:nuts:123?type=NutsCompoundServiceRef` would produce:
+
+```javascript
+{
+  
+  "id": "did:nuts:abc#1",
+  "type": "NutsCompoundServiceRef",
+  "serviceEndpoint": {
+    "oauth": "https://auth.example.com"
+  }
 }
 ```
 
--- a/rfc/rfc006-distributed-registry.md
+++ b/rfc/rfc006-distributed-registry.md
@@ -271,7 +271,7 @@ All data is public knowledge. All considerations from [§10 of did-core](https:/
 ## 4. Services
 
 It is to be expected that each DID subject will add services to the DID Document. 
-Specific services will be specified in their own RFC or Bold specification. 
+Specific services will be specified in their own RFC or Bolt specification. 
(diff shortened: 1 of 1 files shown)


rfc/rfc006-distributed-registry.md
```diff
@@ -270,18 +270,23 @@ All data is public knowledge. All considerations from [§10 of did-core](https:/
 
 ## 4. Services
 
-It is to be expected that each DID subject will add services to the DID Document. Specific services will be specified in their own RFC or Bold specification. A service can define an absolute endpoint URI or be a compound service, referring to a set of services. This is often the case when a SaaS provider defines endpoints to be used for all clients.
+It is to be expected that each DID subject will add services to the DID Document. 
+Specific services will be specified in their own RFC or Bold specification.
```
reinkrul, line 274: Bold -> Bolt

rfc/rfc006-distributed-registry.md
```diff
@@ -270,18 +270,23 @@ All data is public knowledge. All considerations from [§10 of did-core](https:/
...
 
-For an absolute endpoint URI the `serviceEndpoint` MUST be a string containing the URI. For a compound service the `serviceEndpoint` MUST contain a map containing references to absolute endpoint URI services. The references MUST be by query and not by fragment. Only `type` can be used as query param and it refers to the `type` field in a service. A compound service MAY refer to absolute endpoints from other DID Documents. See [§3.2 of did-core](https://www.w3.org/TR/did-core/#did-url-syntax) for DID URL syntax and [RFC3986](https://tools.ietf.org/html/rfc3986) for generic URL standards.
+For an absolute endpoint URI the `serviceEndpoint` MUST be a string containing a URL. 
+For a compound service the `serviceEndpoint` MUST contain a map with absolute endpoint URLs or references to other services. 
+References MUST be by query and not by fragment. Only `type` can be used as query param and it refers to the `type` field in a service. A compound service MAY refer to absolute endpoints from other DID Documents. See [§3.2 of did-core](https://www.w3.org/TR/did-core/#did-url-syntax) for DID URL syntax and [RFC3986](https://tools.ietf.org/html/rfc3986) for generic URL standards.
```
reinkrul, line 281: add an example

rfc/rfc006-distributed-registry.md
```diff
@@ -322,7 +327,66 @@ The care organization refers to it:
 }
 ```
 
-### 4.1. Contact information
+### 4.1 Service references
+
+Any `serviceEndpoint` with a value that starts with `did` is a reference to another service.
+As specified earlier, all references use the `type` query parameter.
+The origin of a reference is from the `serviceEndpoint` context, while the target is a complete `service` object.
```
reinkrul, line 334: rephrase

rfc/rfc006-distributed-registry.md
```diff
@@ -322,7 +327,66 @@ The care organization refers to it:
...
+When a reference targets a `service`, it actually targets the `serviceEndpoint` of the service.
+
+A client doesn't care about the references. When a client queries for particular services, the references need to be resolved.
+Resolving a reference might lead to resolving more references and may lead to an infinite loop.
```
reinkrul, line 338: references, and may lead to an infinite loop

reinkrul, line 338: Let's just say references MUST NOT create a cycle/loop?

rfc/rfc006-distributed-registry.md
```diff
@@ -322,7 +327,66 @@ The care organization refers to it:
...
+Resolving a reference MUST NOT go deeper than 5 levels. Given 5 services, A through E, A references B, B references C, etc. Given the maximum depth of 5, E MUST contain an absolute endpoint URI.
```
reinkrul, line 339: 5 sounds like a lot, maybe just start at 3? We can always up it (lowering the limit is much harder)

Given 5 services where `A` ultimately references `E` through `B`, `C` and `D` (`A -> B -> C -> D -> E`)

rfc/rfc006-distributed-registry.md
```diff
@@ -322,7 +327,66 @@ The care organization refers to it:
...
+
+Another limitation is that a reference that originates from a compound service MUST NOT resolve to another compound service. When resolving that would lead to a map of URIs nested within a map of URIs.
```
reinkrul, line 341: Drop "When"

If you specify the effect, ("resolving that would lead to a map..." etc), specify the problem as well

reinkrul (changes requested):

rfc/rfc006-distributed-registry.md
woutslakhorst: well 3 would be close to a real use-case, so 5 is safe

reinkrul (approved):

woutslakhorst:
to be implemented by nuts-foundation/nuts-node#340

PR #117, opened 2021-07-28
changes to RFCs when ordering is required for KID resolution

woutslakhorst:
This PR changes the `kid` resolution from using `sigt` to using transaction ordering.

In the nuts-node code, the change would be to resolve a DID Document using the hash in the query metadata instead of the timestamp.

Introduction in the code has to be done in 2 steps (or a dev network reset):

- add additional TX refs to `prevs` when publishing
- after a while change resolving of new TXs (after a certain epoch) to use the new resolving.

### DIFF
--- a/rfc/rfc004-verifiable-transactional-graph.md
+++ b/rfc/rfc004-verifiable-transactional-graph.md
@@ -43,7 +43,7 @@ Other terminology comes from the [Nuts Start Architecture](rfc001-nuts-start-arc
 
 Transactions MUST be encoded as [RFC7515 JSON Web Signature](https://tools.ietf.org/html/rfc7515). It MUST be serialized in compact \([RFC7515 section 7.1](https://tools.ietf.org/html/rfc7515#section-7.1)\) serialization form.
 
-### 3.1. JWS implementation
+### 3.1 JWS implementation
 
 In addition to required header parameters as specified in RFC7515 the following requirements apply:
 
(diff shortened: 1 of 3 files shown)


gerardsn (commented):
Comments only since I do not fully understand the impact of this change

rfc/rfc004-verifiable-transactional-graph.md
```diff
@@ -136,11 +136,21 @@ until queue empty; take transaction from queue
...
+Before interpreting a transaction's content the JWS' signature MUST be validated. Almost all transactions will use the **kid** to identify the key used to sign the transaction.
+Resolving the **kid** based on the signature timestamp (**sigt**) can cause problems and even adds attack vectors.
+The main reason behind this, is that the signature time doesn't set the order of transactions.
+Transaction B can be signed with a key introduced in transaction A. Transaction B depends on A, this dependency can't be solved with the signature time.
+This dependency can be solved by simply referring to the transaction that introduced the **kid**.
+If a ``kid`` is present in the JWS, the `prevs` MUST contain a transaction reference of which the transaction content includes the key entry with the **kid** as identifier.
+If the resolved transaction contents is not the latest version, this may indicate a broken installation or a stolen private key.
```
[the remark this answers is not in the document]
woutslakhorst: Indeed, but this RFC doesn't know about DID docs.... chicken and egg problem

rfc/rfc004-verifiable-transactional-graph.md
```diff
@@ -136,15 +136,24 @@ until queue empty; take transaction from queue
     process transaction
 ```
 
-### 3.6. Signature and transaction content verification
+### 3.6 Signature verification
 
-Before interpreting a transaction's content the JWS' signature MUST be validated. When **kid** is used to specify the signing key and the system knows additional usage restrictions \(e.g. the key is valid from X to Y, to be checked against **sigt**\), the system MUST assert the usage is compliant.
+Before interpreting a transaction's content the JWS' signature MUST be validated. Almost all transactions will use the **kid** to identify the key used to sign the transaction.
+Resolving the **kid** based on the signature timestamp (**sigt**) can cause problems and even adds attack vectors.
+The main reason behind this, is that the signature time doesn't set the order of transactions.
```
reinkrul, line 143: set -> determine

rfc/rfc004-verifiable-transactional-graph.md
```diff
@@ -136,15 +136,24 @@ until queue empty; take transaction from queue
...
+Transaction B can be signed with a key introduced in transaction A. Transaction B depends on A, this dependency can't be solved with the signature time.
+This dependency can be solved by referring to the transaction that introduced the **kid**.
+If a ``kid`` is present in the JWS, the `prevs` MUST contain a transaction reference of which the transaction content includes the key entry with the **kid** as identifier.
+If the resolved transaction contents is not the latest version, this may indicate a broken installation or a stolen private key.
```
reinkrul, line 147: add what the application should/could do in that case?

rfc/rfc004-verifiable-transactional-graph.md
```diff
@@ -136,15 +136,24 @@ until queue empty; take transaction from queue
     process transaction
 ```
 
-### 3.6. Signature and transaction content verification
+### 3.6 Signature verification
 
-Before interpreting a transaction's content the JWS' signature MUST be validated. When **kid** is used to specify the signing key and the system knows additional usage restrictions \(e.g. the key is valid from X to Y, to be checked against **sigt**\), the system MUST assert the usage is compliant.
+Before interpreting a transaction's content the JWS' signature MUST be validated. Almost all transactions will use the **kid** to identify the key used to sign the transaction.
+Resolving the **kid** based on the signature timestamp (**sigt**) can cause problems and even adds attack vectors.
```
reinkrul, line 142: I think the **kid** requirement can be part of section 3.1 where (additional) requirements on JWS fields are stated?

reinkrul (changes requested):
Solution sounds OK, some minor comments about form

rfc/rfc004-verifiable-transactional-graph.md
woutslakhorst: 3.1 is more on required fields and this kid requirement is quite lengthy. The entire requirement therefore got its own paragraph

rfc/rfc004-verifiable-transactional-graph.md
woutslakhorst: added some explanation

PR #118, opened 2021-07-28
remove tid/tiv references from rfc006

### LINKED ISSUE #114: RFC006 leftover mentions of tiv and tid
old concepts are no longer used: `tiv` and `tid`

woutslakhorst:
Fixes #114


reinkrul (approved):
Didn't know they were still there?

PR #119, opened 2021-08-04
Update the oauth2 spec based on the changes in the flow

dmeijboom:
Based on the changes in: https://github.com/nuts-foundation/nuts-node/pull/391

### DIFF
--- a/rfc/rfc003-oauth2-authorization.md
+++ b/rfc/rfc003-oauth2-authorization.md
@@ -112,7 +112,7 @@ Each compound service MUST define an `oauth` serviceEndpoint. This endpoint refe
 
 * **iss**: The issuer MUST contain the DID of the actor, thus the care organization making the request.
 * **sub**: The subject MUST contain the DID of the custodian. The custodian's DID could be used to find the relevant consent \(together with the actor and subject\).
-* **vcs**: A list of base64 encoded [Nuts Authorization Credentials](rfc014-authorization-credential.md). Optional
+* **vcs**: A list of [Verifiable Credentials](rfc011-verifiable-credential.md). Optional
 * **purposeOfUse**: A list of desired usages. Corresponds to Bolts. Omitted when empty.   
 * **aud**: As per [RFC7523](https://tools.ietf.org/html/rfc7523), the `aud` MUST be an `oauth` service identifier. That service MUST be an absolute endpoint.
 * **usi**: User identity signature. The token container according to the [Authentication token RFC](rfc002-authentication-token.md). Base64 encoded. Optional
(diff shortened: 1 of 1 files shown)


rfc/rfc003-oauth2-authorization.md
```diff
@@ -235,12 +235,12 @@ The **vcs** field contains the authorization credentials that determine the acce
```
stevenvegt, line 235: `The **vcs** field contains the authorization credentials.`
This must be updated so it explains that it can contain multiple credentials?
Also, currently we only have a authorization credential, so perhaps explain more types can be added in the future and each type will have its own set of rules how en when they should be used?

rfc/rfc003-oauth2-authorization.md
```diff
@@ -235,12 +235,12 @@ The **vcs** field contains the authorization credentials that determine the acce
...
 * The credential `issuer` equals the **sub** field of the JWT.
 * The credential `credentialSubject.id` equals the **iss** field of the JWT.
-* It is valid according to [RFC014](rfc014-authorization-credential.md).
 * The credential proof is valid.
 * The credential has not been revoked.
 * The credential has not expired.
 
-The credentials contain the policy and allowed resources. It MUST be verified that the actor's request conforms to the specified policy and only requests allowed resources.
+If the VerifiableCredential has the `NutsAuthorizationCredential type` it's valid according to [RFC014](rfc014-authorization-credential.md) and contains the policy and allowed resources.
+It MUST be verified that the actor's request conforms to the specified policy and only requests allowed resources.
```
stevenvegt, line 243: Can these be rewritten so it follows the same style as the list:
Something like:
```
If the VerifiableCredential has the `NutsAuthorizationCredential type`:
* it's valid according to [RFC014](rfc014-authorization-credential.md)
* It contains the policy and allowed resources.
* The actor's request conforms to the specified policy and only requests allowed resources.
```

rfc/rfc003-oauth2-authorization.md
dmeijboom: Fixed

rfc/rfc003-oauth2-authorization.md
dmeijboom: Fixed

rfc/rfc003-oauth2-authorization.md
```diff
@@ -229,18 +229,19 @@ The **usi** field in the JWT contains the signed login contract. If present it M
 
 The **aud** field MUST match the identifier of the registered endpoint. This prevents the use of the JWT at any other endpoint. The endpoint reference is used for this. \([RFC7523](https://tools.ietf.org/html/rfc7523#section-3)\). An endpoint identifier contains a DID and a fragment.
 
-**5.2.1.7 Validate authorization credentials**
+**5.2.1.7 Validate verifiable credentials**
 
-The **vcs** field contains the authorization credentials that determine the accessible resource. Each credential must valid according to the following rules:
+The **vcs** field contains the list of verifiable credentials of any type.
+
+If a credential has the `NutsAuthorizationCredential` type it determines the accessible resource and is valid according to the following rules (based on [RFC014](rfc014-authorization-credential.md)):
```
reinkrul, line 236: strange sentence, missing "whether" or something else

reinkrul (changes requested):

reinkrul (approved):

PR #121, opened 2021-08-17
RFC006: Fix incorrect DID Document transaction payload type

reinkrul:
Fixed earlier in Nuts Node but appearantely we forgot to fix it in the specs.


dmeijboom (approved):
LGTM

PR #141, opened 2021-11-02
add limitation to V1 protocol for to header

### LINKED ISSUE #140: RFC005: define behaviour when `to` field is present on TX
required for private VCs, it'll contain the encrypted did of the recipient.

woutslakhorst:
Closes #140 

nodes that only speak the v1 protocol should ignore transaction content when the transaction has the `to` header. This indicates that it's a private VC

### DIFF
--- a/rfc/rfc005-distributed-network-using-grpc.md
+++ b/rfc/rfc005-distributed-network-using-grpc.md
@@ -115,7 +115,10 @@ If the response message exceeds the maximum Protobuf message size (as defined by
 
 ### 5.4. Resolving Transaction Content
 
-When the local node is missing a transaction's content, it SHOULD query the peer that provided the transaction for the content using the `TransactionPayloadQuery` message. The peer MUST respond with the `TransactionPayload` message, providing the actual content in the `data` field. If the peer doesn't have the content the `data` field MUST be left empty.
+When the local node is missing a transaction's content, it SHOULD query the peer that provided the transaction for the content using the `TransactionPayloadQuery` message.
+It MAY NOT query the peer for the content if the transaction contains the `to` field. Transactions with this field can only be queried with higher level protocol versions.
+The peer MUST respond with the `TransactionPayload` message, providing the actual content in the `data` field. If the peer doesn't have the content the `data` field MUST be left empty.
+The peer MUST respond with an empty `data` field if the transaction for the requested content contains a `to` field.
 
 When the transaction's payload can't be resolved from the peer that provided the transaction, the node MAY broadcast the `TransactionPayloadQuery` to resolve it from one of its other peers.
 Nodes MUST avoid broadcasting this message too often, otherwise it might put stress the network.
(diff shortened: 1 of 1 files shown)


rfc/rfc005-distributed-network-using-grpc.md
```diff
@@ -115,7 +115,10 @@ If the response message exceeds the maximum Protobuf message size (as defined by
 
 ### 5.4. Resolving Transaction Content
 
-When the local node is missing a transaction's content, it SHOULD query the peer that provided the transaction for the content using the `TransactionPayloadQuery` message. The peer MUST respond with the `TransactionPayload` message, providing the actual content in the `data` field. If the peer doesn't have the content the `data` field MUST be left empty.
+When the local node is missing a transaction's content, it SHOULD query the peer that provided the transaction for the content using the `TransactionPayloadQuery` message.
+It MAY NOT query the peer for the content if the transaction contains the `to` field. Transactions with this field can only be queried with higher level protocol versions.
+The peer MUST respond with the `TransactionPayload` message, providing the actual content in the `data` field. If the peer doesn't have the content the `data` field MUST be left empty.
+The peer MUST respond with an empty `data` field if the transaction for the requested content contains a `to` field.
```
dmeijboom, line 121: Maybe the `It MAY NOT query (...)` line could come after `The peer MUST respond (...)` line? I think that would make more sense reading the paragraph.

dmeijboom (changes requested):

dmeijboom (approved):
LGTM

rfc/rfc005-distributed-network-using-grpc.md
```diff
@@ -115,7 +115,10 @@ If the response message exceeds the maximum Protobuf message size (as defined by
 
 ### 5.4. Resolving Transaction Content
 
-When the local node is missing a transaction's content, it SHOULD query the peer that provided the transaction for the content using the `TransactionPayloadQuery` message. The peer MUST respond with the `TransactionPayload` message, providing the actual content in the `data` field. If the peer doesn't have the content the `data` field MUST be left empty.
+When the local node is missing a transaction's content, it SHOULD query the peer that provided the transaction for the content using the `TransactionPayloadQuery` message.
+The peer MUST respond with the `TransactionPayload` message, providing the actual content in the `data` field. If the peer doesn't have the content the `data` field MUST be left empty.
+It MAY NOT query the peer for the content if the transaction contains the `to` field. Transactions with this field can only be queried with higher level protocol versions.
```
reinkrul, line 120: What does "It" refer to here (I assume "the local node")? I'd name it explicitly.

> with higher level protocol versions
 "higher level" sounds a bit weird, "newer protocols", or just "higher protocol versions"

If the peer MUST respond with an empty `data` field for these TXs we can say the peer "SHOULD NOT" query it?

reinkrul (changes requested):

rfc/rfc005-distributed-network-using-grpc.md
woutslakhorst: adjusted the first 2. The last point is because evil nodes might still try it and an answer should be defined in that case.

reinkrul (approved):

PR #145, opened 2021-11-08
added rfc016: lamport clocks

woutslakhorst:
required for #126

This PR adds a new transaction version and changes to processing behaviour.

Introduction to the new TX version will have to be 2 fold:
- enable processing of new TXs and apply virtual lamport clocks
- enable creation of new TXs

### DIFF
--- a/rfc/rfc016-lamport-clock.md
+++ b/rfc/rfc016-lamport-clock.md
@@ -1,0 +1,63 @@
+# RFC016 Transaction Lamport clock
+
+|  |  |
+| :--- | :--- |
+| Nuts foundation | W.M. Slakhorst |
+| Request for Comments: 016 | Nedap |
+| Updates: RFC004 | November 2021 |
+
+## Transaction Lamport clock
+
+### Abstract
+
+This RFC updates [RFC004](rfc004-verifiable-transactional-graph.md). It specifies a version two transaction format and adds additional requirements to version 1.
+All additional requirements are backwards compatible. This RFC introduces the concept of the [Lamport clock](https://en.wikipedia.org/wiki/Lamport_timestamp) to the transaction contents.
+
+### Status of document
+
+This document is currently a draft.
+
+### Copyright Notice
+
+![](../.gitbook/assets/license.png)
+
+This document is released under the [Attribution-ShareAlike 4.0 International \(CC BY-SA 4.0\) license](https://creativecommons.org/licenses/by-sa/4.0/).
+
+## 1.  Introduction
+
+Version one of the transaction format introduced an ordening based on predecessors. This creates a strict ordering but does not support the exchange of missing transactions between nodes.
+More efficient set reconciliation methods can be used when transactions can be grouped into ranges. 
+
+## 2. Terminology
+
+* **Lamport clock**: Logical clock creating a casual ordering between transactions.
+
+Other terminology comes from the [Nuts Start Architecture](rfc001-nuts-start-architecture.md#nuts-start-architecture).
+
+## 3. New transaction format
+
+Version 2 transactions remain encoded as JWS: [RFC004 ref](rfc004-verifiable-transactional-graph.md#3-transaction-format).
+
+Changes for the version 2 transactions:
+
+* The **ver** protected header MUST be 2.
+* The **lc** protected header is added. It MUST be a positive number constructed as follows:
+  * if the transaction has no entries in **prevs**, it has a lc value that is equal to **sigt**.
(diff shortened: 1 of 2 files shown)


rfc/rfc016-lamport-clock.md
```diff
@@ -0,0 +1,63 @@
...
+## 3. New transaction format
+
+Version 2 transactions remain encoded as JWS: [RFC004 ref](rfc004-verifiable-transactional-graph.md#3-transaction-format).
+
+Changes for the version 2 transactions:
+
+* The **ver** protected header MUST be 2.
+* The **lc** protected header is added. It MUST be a positive number constructed as follows:
+  * if the transaction has no entries in **prevs**, it has a lc value that is equal to **sigt**.
+  * otherwise, the value is equal to `max(prev1, ... prevN)+1`.
```
dmeijboom, line 46: How do we know it's equal to `max(prev1, ... prevN)+1`? The Lamport timestamp could also be a different value as long as it's more than `max(prev1, ... prevN)`.

rfc/rfc016-lamport-clock.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 3. New transaction format
+
+Version 2 transactions remain encoded as JWS: [RFC004 ref](rfc004-verifiable-transactional-graph.md#3-transaction-format).
+
+Changes for the version 2 transactions:
+
+* The **ver** protected header MUST be 2.
+* The **lc** protected header is added. It MUST be a positive number constructed as follows:
+  * if the transaction has no entries in **prevs**, it has a lc value that is equal to **sigt**.
```
dmeijboom, line 45: How can the `lc` value be equal to the signing time? Or is it the idea that the signing time contains a Lamport timestamp as well? If so it should probably be added to the changes list.

rfc/rfc016-lamport-clock.md
```diff
@@ -0,0 +1,63 @@
...
+Although version 1 transactions do not have a **lc** value in the transaction, implementing software MUST assign a value based on the before mentioned algorithm.
+
+### 4.2 Processing the DAG
+
+[§3.5 Processing the DAG](rfc004-verifiable-transactional-graph.md#35-processing-the-dag)is replaced by this paragraph.
+
+Processing the DAG can be seen as planning tasks required for construction: some tasks can happen in parallel \(laying floors and installing electricity\), some tasks must happen sequentially \(foundation must be poured before building the walls\). This is the same for transactions on the DAG: transactions on a branch MUST be processed sequentially but processing order of parallel branches is unspecified, and before processing a merging transaction all **prevs** \(branches\) must have been processed.
+
+Processing MUST be performed in order of the (virtual) **lc** values. 
+If multiple transactions have the same **lc** value, they MUST be processed in sorted order.
```
dmeijboom, line 62: > If multiple transactions have the same **lc** value (...)

A Lamport clock can't return the same Lamport timestamp (that's exactly the point of the algorithm) so this can't happen.

dmeijboom (changes requested):

woutslakhorst:
@dmeijboom I'll rephrase this RFC to use a logical clock since we have some slight different requirements than the lamport clock functionality offers.

woutslakhorst:
> @dmeijboom I'll rephrase this RFC to use a logical clock since we have some slight different requirements than the lamport clock functionality offers.

I actually understood it correctly. The term timestamp in the original paper and on wikipedia is a bit misleading. It's not an epoch just the value of the logical clock.

rfc/rfc016-lamport-clock.md
woutslakhorst: We will make this a condition with MUST

rfc/rfc016-lamport-clock.md
woutslakhorst: We'll start at 0 ;)

rfc/rfc016-lamport-clock.md
dmeijboom: No longer relevant as the can actually overlap with multiple nodes (as they have their own Lamport clock)

woutslakhorst: This happens when multiple nodes will use their internal clock representation in parallel.

rfc/rfc016-lamport-clock.md
woutslakhorst: I'll add a ref to RFC004 and describe the combined effect of using prevs in the TX and LC value

dmeijboom (approved):
LGTM

PR #151, opened 2021-11-29
describe constraint on user identification

### LINKED ISSUE #109: RFC002: BSN and user identification
state that BSN should not be used for user identification. could be added to 4.4

woutslakhorst:
closes #109 

BSN is mentioned as example, main addition is the legislation context.

### DIFF
--- a/rfc/rfc002-authentication-token.md
+++ b/rfc/rfc002-authentication-token.md
@@ -72,7 +72,6 @@ A verifier must be able to verify the token without contacting external services
 
 The token must hold the correct attributes on the user identity for the specific scope the token is to be used for.
 This data is shared with a different legal context (eg: different organization) and is therefore subject to legislation.
-A citizen service number or social security number for example can not be used.
 
 ### 4.5. Cryptographically signed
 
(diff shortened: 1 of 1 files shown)


rfc/rfc002-authentication-token.md
```diff
@@ -71,10 +71,12 @@ A verifier must be able to verify the token without contacting external services
 ### 4.4. User identification
 
 The token must hold the correct attributes on the user identity for the specific scope the token is to be used for.
+This data is shared with a different legal context (eg: different organization) and is therefore subject to legislation.
+A citizen service number or social security number for example can not be used.
```
reinkrul, line 75: Would it be useful to give an example of something that _can_ be used?

reinkrul (changes requested):

rfc/rfc002-authentication-token.md
woutslakhorst: _it depends_

gerardsn: The IRMA/UZI specifications in `## 7. Supported means` provide useful examples?

woutslakhorst: I don't think this is the place to add to much examples. I can remove the negative example to prevent confusion.

PR #159, opened 2021-12-10
initial set reconciliation protocol

woutslakhorst:
This PR describes how set reconciliation works and the individual steps.
It does not have a link to the protobuf spec.

@gerardsn I described the hashing to the (non-colliding) buckets using the rehashing method of the original paper. If we can't find any theory on using a random number generator, this would work as well?

### DIFF
--- a/rfc/rfc017-distributed-network-grpc-v2.md
+++ b/rfc/rfc017-distributed-network-grpc-v2.md
@@ -58,7 +58,7 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
 
 Certain messages in the protocol are a response to a certain request. To make sure a response corresponds to a certain request, a `conversationID` is added to the request. Response type messages MUST add the same `conversationID` in the response.
 The `conversationID` is scoped to a connection.
-A node is free to choose the form of a `conversationID`.
+A node is free to choose the form of a `conversationID`, it MUST be unique during the lifetime of the connection.
 It MUST be valid for at least 10 seconds and no longer than 30 seconds.
 
 If a node receives a response with a `conversationID`, it MUST match its contents with the original request.
@@ -115,7 +115,7 @@ The list of transaction references MUST be tracked per connection.
 If no new transactions have been added/received or for the first message for a connection, an empty list is sent.
 The list MUST not contain more than 100 transactions.
 This could create a backlog of messages at the sending node's side. 
-A node SHOULD take precautions on keeping this backlog to a minimum, eg it SHOULD reduce the inflow of new transactions.
+A node SHOULD take precautions on keeping this backlog below maximum, eg it SHOULD reduce the inflow of new transactions.
 
 The `LC` value MUST equal the highest Lamport Clock value of all transaction references included in the `XOR` calculation.
 If no transactions are present, an all-zero `XOR` and `LC` of 0 is sent.
@@ -130,7 +130,7 @@ If the values are equal, no further action is required.
 If they are not equal, the transaction list MUST be filtered. All known transaction references MUST be removed.
 The resulting list contains all transaction the node is missing.
 The node MUST send a `TransactionListQuery` message containing the list of missing transaction references.
-This message MUST also add a `conversationID`.
+This message MUST also add a new `conversationID`.
 
 If the resulting list of transactions is empty and the `LC` value in the message is equal to or higher than the highest Lamport Clock value of the node, then the node MUST send a `State` message. See §6.2.1.
 
@@ -138,10 +138,10 @@ If the resulting list of transactions is empty and the `LC` value in the message
 
 When a node receives a `TransactionListQuery` message, it MUST respond with a `TransactionList` message.
 This is a response type message so it MUST include the sent `conversationID`.
-Each transaction in the message MUST have been requested. Transactions that resulted from a `TransactionListQuery` MUST have been present in that message.
+Transactions that resulted from a `TransactionListQuery` MUST have been present in that message.
 Unknown transactions references SHOULD be ignored.
 Transactions that resulted from a `TransactionRangeQuery` MUST have an LC value that is within the requested range.
-When a `TransactionList` message is received and this is not true, the entire message MUST be ignored.
+If any of these requirements are not met, the entire message MUST be ignored.
 
 A `TransactionList` message MAY be broken up into smaller messages, each message should confirm to these rules. Each part MUST also use the same `conversationID`.
 All transactions in the `TransactionList` message MUST be sorted by LC value (lowest first).
@@ -155,12 +155,12 @@ depends on private tx chaptr
 One of the problems in a distributed network is how to make sure every node has processed all transactions.
 When two nodes haven't processed the same set of transactions, the second problem is how to efficiently synchronize the missing transactions.
 The higher the efficiency, the more transactions the entire network can process.
-This part of the protocol is used to synchronize transactions that are missed by the gossip protocol. 
+This part of the protocol is used to synchronize transactions that are missed by the gossip protocol (e.g. due to nodes being offline/network partitions). 
 
 ### 6.1 Data requirements
 
 The protocol requires an XOR value and IBLT structure to be sent in a message.
-The XOR value has already ben explained in §5.1.1.
+The XOR value has already been explained in §5.1.1.
 These are calculated over different transaction ranges based on LC value.
 
 #### 6.1.1 Invertible Bloom Lookup Table
@@ -240,7 +240,7 @@ A node MUST make sure to only add transactions of which all previous transaction
 #### 6.2.1 Broadcasting the state
 
 The `State` message is sent as response to various conditions of the gossip protocol (see §5).
-The `State` message MUST contain a `conversationID`.
+The `State` message MUST contain a new `conversationID`.
 The `State` message contains a `XOR` value and an `LC`.
 The `LC` value MUST equal the highest Lamport Clock value of all transaction references included in the `XOR` calculation.
 If no transactions are present, an all-zero `XOR` and `LC` of 0 is sent.
@@ -311,13 +311,13 @@ The protocol uses a variety of message to synchronize all transactions over node
 Together they cover all situations.
 
 If all nodes are operating correctly, the gossip protocol is the most efficient in synchronizing all the transactions.
-It will not be able to deal with situations where a node is offline or when nodes can no longer reach each other.
-It can be used frequent and uses small messages.
+It will not be able to deal with situations where a node has been offline or when nodes can no longer reach each other.
+It can be used frequently because it uses small messages.
 
 If a node has been offline for some time, the combination of an IBLT and a range query will synchronize the node efficiently.
-The IBLT will resolve all missing transactions for the node's latest page and the range query will synchronize all the missing pages. Those pages will include all transactions that have been created during the period the nodes was offline.
+The IBLT will resolve all missing transactions for the node's latest page and the range query will synchronize all the missing pages. Those pages will include all transactions that have been created during the period the node was offline.
 
-If a node has not been offline, but its communication has been interrupted, it will have created new transactions.
+If a node has not been offline, but its communication has been interrupted, it might have created new transactions that aren't synchronized with the rest of the network.
 The Lamport Clock values will overlap with the transactions of the rest of the network.
 The IBLT will be able to determine missing transactions. 
-If the difference in transactions is to big, the protocol will reduce the range the IBLT covers until an IBLT has been found that can resolve the transaction set difference.
+If the difference in transactions is too big, the protocol will reduce the range the IBLT covers until an IBLT has been found that can resolve the transaction set difference.
--- a/rfc/rfc017-distributed-network-grpc-v2.md
+++ b/rfc/rfc017-distributed-network-grpc-v2.md
@@ -30,8 +30,8 @@ This document is released under the [Attribution-ShareAlike 4.0 International \(
 ## 2. Terminology
  
 * **DAG** \(Directed Acyclic Graph\): graph formed of all transactions. It provides casual ordering for transactions and means to efficiently compare the local DAG with those of peers. See [RFC004](rfc004-verifiable-transactional-graph.md) for details.
-* **Invertible Bloom Lookup Table (IBLT)**: A bloom filter where instead of just a 0 or 1, a count, key_sum and hash_sum is stored per bucket. Original paper: https://arxiv.org/pdf/1101.2245 or https://www.ics.uci.edu/~eppstein/pubs/EppGooUye-SIGCOMM-11.pdf.
-* **Lamport Clock (LC)**: Monotonically increasing clock as defined in [RFC016](rfc016-lamport-clock.md).
+* **Invertible Bloom Lookup Table (IBLT)**: A bloom filter where instead of just a 0 or 1, a count, key_sum and hash_sum is stored per bucket. [Original paper](https://www.ics.uci.edu/~eppstein/pubs/EppGooUye-SIGCOMM-11.pdf).
+* **Lamport Clock (LC)**: Logical clock as defined in [RFC016](rfc016-lamport-clock.md).
 * **Node**: local Nuts software system acting \(a.k.a. _Nuts Node_\).
 * **Node identity**: (a.k.a. node DID) the DID a node uses to identify itself on the Nuts network. Multiple logical nodes (e.g. a cluster) may share the same node identity.
 * **Peer**: remote Nuts node that communicates with the local node.
@@ -56,16 +56,21 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
 
 ## 4. Set reconciliation protocol
 
+One of the problems in a distributed network is how to make sure every node has processed all transactions.
+When two nodes haven't processed the same set of transactions, the second problem is how to efficiently synchronize the missing transactions.
+The higher the efficiency, the more transactions the entire network can process.
+The protocol described here is a compromise on efficiency and fairness. It's able to quickly determine a set of missing transactions, but also doesn't favour any node.
+
+
 ### 4.1 Data requirements
 
 The protocol requires an XOR value and IBLT structure to be sent in a message. 
 These are to be calculated over different transaction ranges based on LC value.
-Both data structures have associative properties allowing for optimized data storage.
 
 #### 4.1.1 XOR
 
-The exclusive-or (XOR) value is calculated over all transactions reference within a range.
-The length of the XOR value is therefore the same as the transaction reference length (SHA256: 32 bytes).
+The exclusive-or (XOR) value is calculated over all transaction references within a range.
+The size of the XOR value is therefore the same as the transaction reference size (SHA256: 32 bytes).
 The XOR value is used to quickly determine if two nodes have processed the exact same set of transactions.
 
 #### 4.1.2 Invertible Bloom Lookup Table
@@ -74,12 +79,14 @@ The IBLT is a data structure capable of finding differences between sets.
 The performance of this data structure is not impacted by the number of entries, but only by the size of the set difference.
 This allows two nodes to compare the entire set of transactions with a relative small data structure.
 
-An IBLT has several parameters that define its characteristics. When comparing two IBLTs, these parameters MUST be the same.
+An IBLT has several parameters that define its characteristics.
+The transaction reference is used as *key*.
+When comparing two IBLTs, these parameters MUST be the same.
 An overview of the parameters:
 
 * **buckets**: defines the size of the IBLT, much like the size of a bloom filter. The larger the size, the bigger the set difference can be. 
-* **Hk**: each key is hashed to select the buckets the key is inserted to. The hash function is applied on the previous output to create extra hash values. This continues until k different values are found. The hash function, its seed and the value for k MUST all be the same for the entire network.
-* **Hc**: each key is hashed to a checksum which is used to validate the inserted key during the deconstruction phase.
+* **Hk**: each key is hashed to select the buckets the key is inserted to. The hash function is applied recursively to create extra hash values. This continues until k different values are found. The hash function, its seed and the value for `k` MUST all be the same for the entire network.
+* **Hc**: each key is hashed to a checksum which is used to validate the inserted key during the decode phase. The hash function and its seed MUST all be the same for the entire network.
 * **data types**: the data types for *count*, *val_sum* and *hash_sum* MUST match ((un)signed, byte-size).
 
 The following parameters are used:
@@ -87,21 +94,27 @@ The following parameters are used:
 * **buckets**: 1024
 * **Hc**: [murmur3](https://en.wikipedia.org/wiki/MurmurHash), 8 bytes, seed: 0x00
 * **Hk**: murmur3, 4 bytes, seed: 0x01, k = 6 
-* **key sum size**: 32 bytes
-* **bucket count field**: 4 bytes
+* **val_sum field size**: 32 bytes
+* **count field size**: 4 bytes
 
 See Appendix A.1 for an explanation.
 
-### 4.1.3 IBLT page size
+#### 4.1.3 IBLT page size
 
 Calculating an IBLT is relatively cheap, but it grows linearly with the number of transactions. This would harm performance of larger networks.
 The immutable nature of transactions make it possible to store intermediate IBLTs and use those as an optimization.
 The LC range parameters of the protocol should be adjusted accordingly to optimize use of these intermediate IBLTs.
 Therefore, they should not be chosen freely.
 
-Any LC value used in the protocol MUST be a multiple of 512 (minus 1, because the LC is 0 based). 
-The use of inclusive or exclusive range parameters must make sure that `end-start` is a multiple of 512.
-In the context of an IBLT this size will be referred to as the `IBLT page size` or `PSIZE` for short.
+To simplify the description of the protocol we introduce a new term: **page**.
+A page contains transaction references for a range of LC values. Each page is a multiple of 512.
+The first page includes transactions with LC values between 0 (inclusive) and 512 (exclusive).
+When a *next* or *previous* page is mentioned, this means 512 has to be added/reduced from the `begin` and `end` values of an LC range.
+
+#### 4.1.4 IBLT decoding
+
+When decoding an IBLT as described in [the original paper](https://arxiv.org/pdf/1101.2245 or https://www.ics.uci.edu/~eppstein/pubs/EppGooUye-SIGCOMM-11.pdf), the result is two sets of values: those that are missing in set A and those that are missing in set B. For this protocol only one of those sets is of interest, the other can be ignored.
+This also makes it more likely the IBLT can be decoded.
 
 ### 4.2 Operation
 
@@ -118,13 +131,13 @@ The protocol generally operates as follows:
     * When the XOR is the same and the LC value equals BOB's highest Lamport Clock value: no action required.
     * When different, send a message containing:
       * LC value sent by Alice
-      * the IBLT that includes the Lamport Clock range of 0-LC
+      * the IBLT that includes the Lamport Clock range of 0-LC(Alice)
       * Bob's highest Lamport Clock value
 
 3. When receiving the message, Alice subtracts Bob's IBLT from her own IBLT for the given range (§4.2.3):
-   * If not deconstructable, go to 1 and send values based on a lower Lamport Clock value.
-   * If deconstructable, send a request for missing transactions
-   * If Bob's highest Lamport Clock value is higher that the LC value sent in 1, then request transactions by Lamport Clock value:
+   * If not decodable, go to 1 and send values based on the previous page.
+   * If decodable, send a request for missing transactions
+   * If Bob's highest Lamport Clock value is higher than the LC value sent in 1, then request transactions by Lamport Clock value:
      * Lamport Clock start value.
      * Lamport Clock end value.
 
@@ -136,47 +149,51 @@ A node MUST make sure to only add transactions of which all previous transaction
 
 The sections below specify the details of the protocol operation and maps it to the gRPC messages. See the section "Protobuf Definition" for a full specification of the protobuf/gRPC contract.
 
-### 4.2.1 Broadcasting
+#### 4.2.1 Broadcasting the state
 
-The local node's DAG state MUST be broadcast at an interval using the `AdvertState` message, by default every 2 seconds. The interval MAY be adjusted by the node operator but MUST conform to the limits \(min/max interval\) defined by the network. It is advised to keep it relatively short because it directly influences the speed by which new transactions are retrieved.
+todo:
+The local node's DAG state MUST be broadcast at an interval using the `State` message, by default every 2 seconds. The interval MAY be adjusted by the node operator but MUST conform to the limits \(min/max interval\) defined by the network. It is advised to keep it relatively short because it directly influences the speed by which new transactions are retrieved.
 
-The `AdvertState` message contains a `XOR` value, an `LC` value and a `token`. The `token` is scoped to a connection and is required in the response from the peer.
+The `State` message contains a `XOR` value, an `LC` value and a `token`. The `token` is scoped to a connection and is required in the response from the peer.
 The `LC` value MUST equals the highest Lamport Clock value of all transaction references included in the `XOR` calculation.
 If no transactions are present, an all-zero `XOR` and `LC` of 0 is sent.
 A node is free to choose how the `token` is created as long as it's able to satisfy the requirements of §4.2.3.
 It MUST be valid for at least 10 seconds.
 
-### 4.2.2 IBLT response
+#### 4.2.2 IBLT response
 
-When a peer receives an `AdvertState` message and the `XOR` differs, it SHOULD respond with a `TransactionSet` message.
-The upper bound of the response IBLT MUST be the equal to `PSIZE*ceil(LC/PSIZE)` where `LC` is the value from the `AdvertState` message and `ceil` is a function that rounds a double up to the nearest integer.
-If `LC` equals zero, `PSIZE` is used.
-This upper bound is exclusive, the lower bound is inclusive and always 0.
-Next to the IBLT data, the `TransactionSet` message MUST contain the original `LC` value as `LC_req`, a new `LC` value indicating the highest LC value from the peer and the `token` value from the `AdvertState` message.
+When a peer receives an `State` message and the `XOR` differs, it SHOULD respond with a `TransactionSet` message.
+The sent `LC` value falls within the bounds of a page.
+The response IBLT MUST be calculated over the transactions leading up to and including that page.
+Next to the IBLT data, the `TransactionSet` message MUST contain the original `LC` value as `LC_req`, a new `LC` value indicating the highest LC value from the peer and the `token` value from the `State` message.
 
-### 4.2.3 Transaction List Query
+#### 4.2.3 Transaction List Query
 
 For every `TransactionSet` message a node receives, it MUST check the `token` value in the message and compare it to its own administration.
 If the `token` doesn't exist or if the `LC_req` value doesn't match, the node MUST ignore the message.
 
-The IBLT in the `TransactionSet` message contains every transaction of the peer in the LC range of `0-LC_req`. The node MUST lookup/compute the IBLT for this range and subtract it from the IBLT of the peer. Deconstruction of the resulting IBLT will list the transaction refs the peer has and the local node misses. 
+The IBLT in the `TransactionSet` message contains every transaction of the peer in the LC range of `0-LC_req`. The node MUST lookup/compute the IBLT for this range and subtract it from the IBLT of the peer. Decoding the resulting IBLT will list the transaction refs the peer has and the local node misses. 
 
 These transactions MUST be queried by using the `TransactionListQuery` message. It contains all missing transaction references.
 The message MUST contain a `token` that is to be used in the response.
 
-If the deconstruction fails, the local node sends a new `AdvertState` message. The `LC` value MUST be `PSIZE*floor((LC_req-1)/PSIZE)` where `floor` is a function that rounds a double down to the nearest integer. 
+If the decoding fails, the local node sends a new `State` message. The `LC` value MUST be the `end` of the page before the page that included `LC_req`. 
 The `XOR` value MUST be calculated over the transaction references in the range starting with 0 and ending with the new `LC` value.
 
-### 4.2.4 Transaction Range Query
+#### 4.2.4 Transaction Range Query
 
-If the IBLT from a `TransactionSet` message could be deconstructed, the local node MAY also send a `TransactionRangeQuery` message.
-This is only needed if the `LC` value of the `TransactionSet` message is equal or larger than `PSIZE*ceil(LC_req/PSIZE)`. 
+If the IBLT from a `TransactionSet` message could be decoded, the local node MAY also send a `TransactionRangeQuery` message.
+This is only needed if the page containing the `LC` value of the `TransactionSet` message comes after the page containing `LC_req`. 
 This means the peer has additional transactions outside the IBLT range.
-The `TransactionRangeQuery` message contains a `begin` and `end` parameter indicating a request for transactions which LC value is between those parameters.
-The `begin` MUST be set to `PSIZE*ceil(LC_req/PSIZE)` and the end MUST be set to `LC`. Both are inclusive.
+If the `LC_req` value is in the latest page of the local node, it SHOULD query all pages leading up to the `LC` value.
+If the `LC_req` value is NOT in the latest page of the local node, then it MUST only query the next page.
+
+The `TransactionRangeQuery` message contains a `start` (inclusive) and `end` (exclusive) parameter corresponding to that requested page(s).
 The message MUST contain a `token` that is to be used in the response.
 
-### 4.2.5 Transaction List
+If decoding fails and the IBLT covered the first page, the local node MUST query the first page.
+
+#### 4.2.5 Transaction List
 
 The `TransactionList` message contains a list of transactions and the `token` from either the `TransactionListQuery` or `TransactionRangeQuery`.
 Each transaction in the message MUST relate to the original request. Transactions that resulted from a `TransactionListQuery` MUST have been present in that message.
@@ -184,6 +201,10 @@ Transactions that resulted from a `TransactionRangeQuery` MUST have an LC value 
 If this is not true, the entire message MUST be ignored.
 A `TransactionList` message MAY be broken up into smaller messages, each message should confirm to these rules. Each part MUST also use the same `token`.
 All transactions in the `TransactionList` message MUST be sorted by LC value (lowest first).
+
+## 5 Gossip
+
+coming
 
 ## X. Private Transactions
 
@@ -199,9 +220,26 @@ In the protocol described here, this would mean that an entire range of transact
 When this happens randomly, it wouldn't be such a problem, but it's also an attack vector.
 An attacker could craft pairs of transactions that would trigger other nodes into requesting large amounts of transactions. 
 This type of attack is called the [birthday attack](https://en.wikipedia.org/wiki/Birthday_attack).
-The only way to protect against this type of attack is to make the set large enough that the cost to the attacker would become to big.
-For an IBLT with 64 buckets and 4 hashes, the change for a collision is 50% after adding ~5000 keys.
+The only way to protect against this type of attack is to make the set large enough that the cost to the attacker would become to high.
+For an IBLT with 64 buckets and 4 hashes, the chance for a collision is 50% after adding ~5000 keys.
 This would be trivial for current computers.
 Expanding this to 1024 buckets and 6 hashes, reaching a 50% chance for a collision is only achieved after adding more than a billion keys.
 This is still not much, but the keys used for the collision have to be created by real signed transactions.
 Even on modern hardware, this could take an hour.
+
+### A.2 The right message for the right situation
+
+The protocol uses a variety of message to synchronize all transactions over nodes.
+Together they cover all known situations.
+
+If all nodes are operating correctly, the gossip protocol is the most efficient in synchronizing all the transactions.
+It will not be able to deal with situations where a node is offline or when nodes can no longer reach each other.
+It can be used frequent and uses small messages.
+
+If a node has been offline for some time, the combination of an IBLT and a range query will synchronize the node efficiently.
+The IBLT will resolve all missing transactions for the node's latest page and the range query will sycnhronize all the missing pages. Those pages will include all transactions that have been created during the period the nodes was offline.
+
+If a node has not been offline, but its communication has been interrupted, it will have created new transactions.
+The Lamport Clock values will overlap with the transactions of the rest of the network.
+The IBLT will be able to determine missing transactions. 
+If the difference in transactions is to big, the protocol will reduce the range the IBLT covers until an IBLT has been found that can resolve the transaction set difference.
--- a/rfc/rfc017-distributed-network-grpc-v2.md
+++ b/rfc/rfc017-distributed-network-grpc-v2.md
@@ -196,8 +196,8 @@ When this happens randomly, it wouldn't be such a problem, but it's also an atta
 An attacker could craft pairs of transactions that would trigger other nodes into requesting large amounts of transactions. 
 This type of attack is called the [birthday attack](https://en.wikipedia.org/wiki/Birthday_attack).
 The only way to protect against this type of attack is to make the set large enough that the cost to the attacker would become to big.
-For an IBLT with 64 buckets and 4 hashes, the change for a collision is 50% after adding ~5133 keys.
+For an IBLT with 64 buckets and 4 hashes, the change for a collision is 50% after adding ~5000 keys.
 This would be trivial for current computers.
-Expanding this to 1024 buckets and 5 hashes, reaching a 50% chance for a collision is only achieved after adding more than 42 million keys.
+Expanding this to 1024 buckets and 5 hashes, reaching a 50% chance for a collision is only achieved after adding more than 40 million keys.
 This is still not much, but the keys used for the collision have to be created by real signed transactions.
-Even on modern hardware, 42 million signing operations would take minutes. Even a 6th hash can be added later to multiply that amount by 32.
+Even on modern hardware, 40 million signing operations would take minutes. A 6th hash can be added later to require more than 1 billion keys for a 50% chance of collision.
--- a/rfc/rfc017-distributed-network-grpc-v2.md
+++ b/rfc/rfc017-distributed-network-grpc-v2.md
@@ -165,7 +165,7 @@ The `XOR` value MUST be calculated over the transaction references in the range 
 
 ### 4.2.4 Transaction Range Query
 
-If a `TransactionSet` message results in a `TransactionListQuery` message, the local node MAY also send a `TransactionRangeQuery` message.
+If the IBLT from a `TransactionSet` message could be deconstructed, the local node MAY also send a `TransactionRangeQuery` message.
 This is only needed if the `LC` value of the `TransactionSet` message is equal or larger than `512*ceil(LC_req/512)`. 
 This means the peer has additional transactions outside the IBLT range.
 The `TransactionRangeQuery` message contains a `begin` and `end` parameter indicating a request for transactions which LC value is between those parameters.
--- a/rfc/rfc017-distributed-network-grpc-v2.md
+++ b/rfc/rfc017-distributed-network-grpc-v2.md
@@ -143,7 +143,7 @@ Unknown transactions references SHOULD be ignored.
 Transactions that resulted from a `TransactionRangeQuery` MUST have an LC value that is within the requested range.
 If any of these requirements are not met, the entire message MUST be ignored.
 
-A `TransactionList` message MAY be broken up into smaller messages, each message should confirm to these rules. Each part MUST also use the same `conversationID`.
+A `TransactionList` message MAY be broken up into smaller messages, each message should conform to these rules. Each part MUST also use the same `conversationID`.
 All transactions in the `TransactionList` message MUST be sorted by LC value (lowest first).
 
 #### 5.2.4 Transaction Payload query
--- a/rfc/rfc017-distributed-network-grpc-v2.md
+++ b/rfc/rfc017-distributed-network-grpc-v2.md
@@ -1,10 +1,12 @@
 # RFC017 Distributed Network Protocol (v2) using gRPC
 
-|  |  |
-| :--- | :--- |
-| Nuts foundation | R.G. Krul |
-| Request for Comments: 017 |  |
-| Replaces: RFC005 | December 2021 |
+|                           |                |
+|:--------------------------|:---------------|
+| Nuts foundation           | R.G. Krul      |
+| Request for Comments: 017 | Nedap          |
+| Replaces: RFC005          | W.M. Slakhorst |
+|                           | Nedap          |
+|                           | December 2021  |
 
 ## Distributed Network Protocol (v2) using gRPC
 
@@ -26,14 +28,21 @@ This document is released under the [Attribution-ShareAlike 4.0 International \(
 
 
 ## 2. Terminology
-
-* **peer**: remote Nuts node that communicates with the local node.
-* **node identity**: (a.k.a. node DID) the DID a node uses to identify itself on the Nuts network. Multiple logical nodes (e.g. a cluster) may share the same node identity.
-* **private transactions**: transactions that are meant for a specific receiver (node) on the network. The transaction payload is solely shared with that particular node.
+ 
+* **DAG** \(Directed Acyclic Graph\): graph formed of all transactions that. It provides casual ordering for transactions and means to efficiently compare the local DAG with those of peers.
+* **Invertible Bloom Lookup Table (IBLT)**: A bloom filter where instead of just a 0 or 1, a count, key_sum and hash_sum is stored per bucket. Original paper: https://arxiv.org/pdf/1101.2245 or https://www.ics.uci.edu/~eppstein/pubs/EppGooUye-SIGCOMM-11.pdf.
+* **Lamport Clock (LC)**: Monotonically increasing clock as defined in [RFC016](rfc016-lamport-clock.md).
+* **Node**: local Nuts software system acting \(a.k.a. _Nuts Node_\).
+* **Node identity**: (a.k.a. node DID) the DID a node uses to identify itself on the Nuts network. Multiple logical nodes (e.g. a cluster) may share the same node identity.
+* **Peer**: remote Nuts node that communicates with the local node.
+* **Private transactions**: transactions that are meant for a specific receiver (node) on the network. The transaction payload is solely shared with that particular node.
+* **Set reconciliation**: process of finding the differences in two sets of data entries and using the least amount of possible transfers to synchronize the sets.
+* **Transaction**: self-contained unit of application data on the DAG.
+* **Transaction reference**: SHA256 of the transaction data.
 
 Other terminology comes from the [Nuts Start Architecture](rfc001-nuts-start-architecture.md#nuts-start-architecture).
 
-## X. Authenticating node identity
+## 3. Authenticating node identity
 
 Particular exchanges (private transactions) might require authentication of the peer's identity. This identity (a.k.a. node identity) is specified by [RFC015 Node identity](rfc015-node-identity.md).
 If a node has a node DID and wishes create an authenticated connection, it MUST send the DID as `nodeDID` gRPC header when establishing inbound or outbound connections.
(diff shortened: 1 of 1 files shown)


rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+#### 4.1.2 Invertible Bloom Lookup Table
+
+The IBLT is a data structure capable of finding differences in sets.
+The performance of this data structure is not impacted by the number of entries, but only by the size of the set difference.
+This allows two nodes to compare the entire set of transactions with a relative small data structure.
+
+An IBLT has several parameters that define its characteristics. When comparing two IBLTs, these parameters MUST be the same.
+An overview of the parameters:
+
+* **#buckets**: defines the size of the IBLT, much like the size of a bloom filter. The larger the size, the bigger the set difference can be.
```
gerardsn, line 80:
```suggestion
* **buckets**: defines the size of the IBLT, much like the size of a bloom filter. The larger the size, the bigger the set difference can be. 
```

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+The IBLT is a data structure capable of finding differences in sets.
+The performance of this data structure is not impacted by the number of entries, but only by the size of the set difference.
+This allows two nodes to compare the entire set of transactions with a relative small data structure.
+
+An IBLT has several parameters that define its characteristics. When comparing two IBLTs, these parameters MUST be the same.
+An overview of the parameters:
+
+* **#buckets**: defines the size of the IBLT, much like the size of a bloom filter. The larger the size, the bigger the set difference can be. 
+* **Hk**: each key is hashed to select the buckets the key is inserted to. The hash function is applied on the previous output to create extra hash values. This continues until k different values are found. The hash function, its seed and the value for k MUST all be the same for the entire network.
```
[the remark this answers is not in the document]
gerardsn: detail: the hashing function produces numbers much larger than the number of buckets. Since 2 different hashes can map to the same bucket, we need to keep hashing until we find k different buckets not values.

gerardsn (changes requested):
Looks good
- main question: the protocol appears to prefer syncing historical additions before recent (high LC) additions.
- some of the terminology is confusing, especially node vs peer when the definition of node/peer changes every step of the protocol.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+The message MUST contain a `token` that is to be used in the response.
+
+### 4.2.5 Transaction List
+
+The `TransactionList` message contains a list of transactions and the `token` from either the `TransactionListQuery` or `TransactionRangeQuery`.
+Each transaction in the message MUST relate to the original request. Transactions that resulted from a `TransactionListQuery` MUST have been present in that message.
+Transactions that resulted from a `TransactionRangeQuery` MUST have an LC value that is within the requested range.
+If this is not true, the entire message MUST be ignored.
+A `TransactionList` message MAY be broken up into smaller messages, each message should confirm to these rules. Each part MUST also use the same `token`.
+All transactions in the `TransactionList` message MUST be sorted by LC value (lowest first).
```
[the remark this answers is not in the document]
woutslakhorst: from https://grpc.io/docs/what-is-grpc/core-concepts/

_Bidirectional streaming RPCs where both sides send a sequence of messages using a read-write stream. The two streams operate independently, so clients and servers can read and write in whatever order they like: for example, the server could wait to receive all the client messages before writing its responses, or it could alternately read a message then write a message, or some other combination of reads and writes. The order of messages in each stream is preserved._

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+The IBLT in the `TransactionSet` message contains every transaction of the peer in the LC range of `0-LC_req`. The node MUST lookup/compute the IBLT for this range and subtract it from the IBLT of the peer. Deconstruction of the resulting IBLT will list the transaction refs the peer has and the local node misses. 
+
+These transactions MUST be queried by using the `TransactionListQuery` message. It contains all missing transaction references. An IBLT deconstruction with the current parameters will not be able to yield anything larger than ~750 values. This will easily fit into a single message given the current maximum message size.
+The message MUST contain a `token` that is to be used in the response.
+
+If the deconstruction fails, the local node sends a new `AdvertState` message. The `LC` value MUST be `512*floor((LC_req-1)/512)` where `floor` is a function that rounds a double down to the nearest integer. 
+The `XOR` value MUST be calculated over the transaction references in the range starting with 0 and ending with the new `LC` value.
+
+### 4.2.4 Transaction Range Query
```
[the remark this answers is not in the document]
woutslakhorst: The idea fow now is that new nodes will use the same protocol, which will result in a single range query from 0-end.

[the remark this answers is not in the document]
woutslakhorst: I did forget to add what to send as XOR/IBLT when no transactions are available!

rfc/rfc017-distributed-network-grpc-v2.md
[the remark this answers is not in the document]
woutslakhorst: correct, as described in the paper.
If a hash function is uniform, doesn't that automatically mean that hashing twice is uniform?

woutslakhorst:
> main question: the protocol appears to prefer syncing historical additions before recent (high LC) additions.
It actually does the opposite:

- if the first IBLT is deconstructable then it doesn't matter.
- going back in LC range requires more messages between two peers and is therefore slower than communicating with another peer that is up2date. Alice and Bob are communicating and can't deconstruct the IBLT. They go back 5120 LC values. This requires 20 messages, each 50ms => takes a full second to get older transactions. Alice is also communicating with Eve, they don't have to go back in LC value. This requires just 2 messages of 50ms each => synced in 100ms!

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+    * XOR of all known transaction references.
+    * Highest Lamport Clock value over all transactions (LC).
+    
+2. When receiving a peer's broadcast, compare the XOR value:
+    * When the XOR is the same: no action required.
+    * When different, send a message containing
+      * If the LC is incompatible with its own registration: no action required.
+      * the IBLT that includes the Lamport Clock range of 0-LC
+      * the peer LC value
+      * the own highest Lamport Clock value
```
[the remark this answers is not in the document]
woutslakhorst: I'll use Alice and Bob

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+The protocol generally operates as follows:
+
+1. The local node broadcasts at a set interval:
+    * XOR of all known transaction references.
+    * Highest Lamport Clock value over all transactions (LC).
+    
+2. When receiving a peer's broadcast, compare the XOR value:
+    * When the XOR is the same: no action required.
+    * When different, send a message containing
+      * If the LC is incompatible with its own registration: no action required.
```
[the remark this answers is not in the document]
woutslakhorst: removed it, was out of place

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+The performance of this data structure is not impacted by the number of entries, but only by the size of the set difference.
+This allows two nodes to compare the entire set of transactions with a relative small data structure.
+
+An IBLT has several parameters that define its characteristics. When comparing two IBLTs, these parameters MUST be the same.
+An overview of the parameters:
+
+* **#buckets**: defines the size of the IBLT, much like the size of a bloom filter. The larger the size, the bigger the set difference can be. 
+* **Hk**: each key is hashed to select the buckets the key is inserted to. The hash function is applied on the previous output to create extra hash values. This continues until k different values are found. The hash function, its seed and the value for k MUST all be the same for the entire network.
+* **Hc**: each key is hashed to a checksum which is used to validate the inserted key during the deconstruction phase. 
+* **data types**: the data types for *count*, *val_sum* and *hash_sum* MUST match ((un)signed, byte-size).
```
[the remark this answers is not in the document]
woutslakhorst: it's a summary of IBLT characteristics that MUST be the same when comparing two IBLTs

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+### 4.1.3 IBLT page size
+
+Calculating an IBLT is relatively cheap, but it grows linearly with the number of transactions. This would harm performance of larger networks.
+The immutable nature of transactions make it possible to store intermediate IBLTs and use those as an optimization.
+The LC range parameters of the protocol should be adjusted accordingly to optimize use of these intermediate IBLTs.
+Therefore, they should not be chosen freely.
+
+Any LC value used in the protocol MUST be a multiple of 512 (minus 1, because the LC is 0 based). 
+The use of inclusive or exclusive range parameters must make sure that `end-start` is a multiple of 512.
+In the context of an IBLT this size will be referred to as the `IBLT page size`.
```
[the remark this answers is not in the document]
woutslakhorst: as addition or replacement?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+2. When receiving a peer's broadcast, compare the XOR value:
+    * When the XOR is the same: no action required.
+    * When different, send a message containing
+      * If the LC is incompatible with its own registration: no action required.
+      * the IBLT that includes the Lamport Clock range of 0-LC
+      * the peer LC value
+      * the own highest Lamport Clock value
+
+3. When receiving the message, the local node subtracts the given IBLT from its own IBLT for the given range:
+   * If not deconstructable, go to 1 and send values based on a smaller Lamport Clock value.
```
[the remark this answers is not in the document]
woutslakhorst: lower sounds better

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+   * If the peers highest Lamport Clock value is higher that the LC value sent in 1, then request transactions by Lamport Clock value:
+     * Lamport Clock start value.
+     * Lamport Clock end value.
+
+4. When receiving a request for transactions, respond with a message including the requested transactions.
+
+5. After adding a peer's transaction to the DAG \(making sure its cryptographic signature is valid\), query the payload if it's missing.
+
+The node MUST make sure to only add transactions of which all previous transactions are present. The node MUST also add the transaction atomically.
+If one transaction is invalid, ignore the entire message.
```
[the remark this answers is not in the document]
woutslakhorst: I'll remove it for now. It's part of another feature.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+The sections below specify the details of the protocol operation and maps it to the gRPC messages. See the section "Protobuf Definition" for a full specification of the protobuf/gRPC contract.
```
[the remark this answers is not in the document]
woutslakhorst: Not there yet, different PR

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+### 4.2.1 Broadcasting
+
+The local node's DAG state MUST be broadcast at an interval using the `AdvertState` message, by default every 2 seconds. The interval MAY be adjusted by the node operator but MUST conform to the limits \(min/max interval\) defined by the network. It is advised to keep it relatively short because it directly influences the speed by which new transactions are retrieved.
+
+The `AdvertState` message contains a `XOR` value, a `LC` value and a `token`. The `token` is scoped to a connection and is required in the response from the peer.
+A node is free to choose how the `token` is created as long as it's able to satisfy the requirements of the next paragraph.
```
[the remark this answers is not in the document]
woutslakhorst: should be 4.2.3

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+### A.1 IBLT parameters
+
+For a small network, 1024 buckets and 5 hashes is quite heavy. 
+For a nation-wide network this would be reasonable.
+The reason for these parameters is the chance of a collision: a different transaction that results in the exact same buckets as another transaction.
+If there's a collision, it would not be possible to deconstruct an IBLT since there'll be too many buckets that still have a count of 2.
+In the protocol this would mean an entire range of transactions would have to be sent.
+When this happens randomly, it wouldn't be such a problem, but it's also an attack vector.
+An attacker could craft pairs of transactions that would trigger other nodes into requesting large amounts of transactions.
```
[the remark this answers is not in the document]
woutslakhorst: It'll only block direct IBLT comparison and will fall back on sending the entire page. So nothing blocking, just slower. On the other hand, adding the 6th hash is not a very big problem.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+1. The local node broadcasts at a set interval:
+    * XOR of all known transaction references.
+    * Highest Lamport Clock value over all transactions (LC).
+    
+2. When receiving a peer's broadcast, compare the XOR value:
+    * When the XOR is the same: no action required.
+    * When different, send a message containing
+      * If the LC is incompatible with its own registration: no action required.
+      * the IBLT that includes the Lamport Clock range of 0-LC
```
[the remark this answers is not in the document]
woutslakhorst: changed order of items so it comes after the notion of ALice's LC

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+      * the peer LC value
+      * the own highest Lamport Clock value
+
+3. When receiving the message, the local node subtracts the given IBLT from its own IBLT for the given range:
+   * If not deconstructable, go to 1 and send values based on a smaller Lamport Clock value.
+   * If deconstructable, send a request for missing transactions
+   * If the peers highest Lamport Clock value is higher that the LC value sent in 1, then request transactions by Lamport Clock value:
+     * Lamport Clock start value.
+     * Lamport Clock end value.
```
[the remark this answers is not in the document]
woutslakhorst: Yes, when you're missing 1024 transactions, you'll never be able to do this via IBLT, the only option left would be to request a range. If you start with LC=0, Bob will send its IBLT for 0-512 and its LC of 2048. Alice will not be able to deconstruct Bob's IBLT (range had 1024 TXs), then Alice can only fall back on a range query.

The ignore part has been removed for now.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+For every `TransactionSet` message a node receives, it MUST check the `token` value in the message and compare it to its own administration.
+If the `token` doesn't exist or if the `LC_req` value doesn't match, the node MUST ignore the message.
+
+The IBLT in the `TransactionSet` message contains every transaction of the peer in the LC range of `0-LC_req`. The node MUST lookup/compute the IBLT for this range and subtract it from the IBLT of the peer. Deconstruction of the resulting IBLT will list the transaction refs the peer has and the local node misses. 
+
+These transactions MUST be queried by using the `TransactionListQuery` message. It contains all missing transaction references. An IBLT deconstruction with the current parameters will not be able to yield anything larger than ~750 values. This will easily fit into a single message given the current maximum message size.
+The message MUST contain a `token` that is to be used in the response.
+
+If the deconstruction fails, the local node sends a new `AdvertState` message. The `LC` value MUST be `512*floor((LC_req-1)/512)` where `floor` is a function that rounds a double down to the nearest integer.
```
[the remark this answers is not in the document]
woutslakhorst: the `-1` makes sure the floor func works.
If it was 1024, it becomes 1023/512 => 1.99 => 1
This will also works for any arbitrary initial value (highes LC value)

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+The `AdvertState` message contains a `XOR` value, a `LC` value and a `token`. The `token` is scoped to a connection and is required in the response from the peer.
+A node is free to choose how the `token` is created as long as it's able to satisfy the requirements of the next paragraph.
+It MUST be valid for at least 10 seconds.
+
+### 4.2.2 IBLT response
+
+When a peer receives an `AdvertState` message and the `XOR` differs, it SHOULD respond with a `TransactionSet` message.
+The upper bound of the response IBLT MUST be the equal to `512*ceil(LC/512)` where `LC` is the value from the `AdvertState` message and `ceil` is a function that rounds a double up to the nearest integer.
+This upper bound is exclusive, the lower bound is inclusive and always 0.
+Next to the IBLT data, the `TransactionSet` message contains the original `LC` value as `LC_req`, a new `LC` value indicating the highest LC value the peer has and the `token` value from the `AdvertState` message.
```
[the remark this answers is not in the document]
woutslakhorst: any ideas?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+The `AdvertState` message contains a `XOR` value, a `LC` value and a `token`. The `token` is scoped to a connection and is required in the response from the peer.
+A node is free to choose how the `token` is created as long as it's able to satisfy the requirements of the next paragraph.
+It MUST be valid for at least 10 seconds.
+
+### 4.2.2 IBLT response
+
+When a peer receives an `AdvertState` message and the `XOR` differs, it SHOULD respond with a `TransactionSet` message.
+The upper bound of the response IBLT MUST be the equal to `512*ceil(LC/512)` where `LC` is the value from the `AdvertState` message and `ceil` is a function that rounds a double up to the nearest integer.
+This upper bound is exclusive, the lower bound is inclusive and always 0.
```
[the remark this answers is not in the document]
woutslakhorst: It MUST always start at zero, since new TXs in higher ranges may depend on TXs in the lower ranges.

Other rules require ordering: did updates, vc issuance so that would prevent behaving nodes from adding everything to the root. New DIDs can all be added to the root, but given this protocol, what would be the harm? (Massive branching no longer is a problem because we're not publishing heads anymore)

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+Next to the IBLT data, the `TransactionSet` message contains the original `LC` value as `LC_req`, a new `LC` value indicating the highest LC value the peer has and the `token` value from the `AdvertState` message.
+
+### 4.2.3 Transaction List Query
+
+For every `TransactionSet` message a node receives, it MUST check the `token` value in the message and compare it to its own administration.
+If the `token` doesn't exist or if the `LC_req` value doesn't match, the node MUST ignore the message.
+
+The IBLT in the `TransactionSet` message contains every transaction of the peer in the LC range of `0-LC_req`. The node MUST lookup/compute the IBLT for this range and subtract it from the IBLT of the peer. Deconstruction of the resulting IBLT will list the transaction refs the peer has and the local node misses. 
+
+These transactions MUST be queried by using the `TransactionListQuery` message. It contains all missing transaction references. An IBLT deconstruction with the current parameters will not be able to yield anything larger than ~750 values. This will easily fit into a single message given the current maximum message size.
```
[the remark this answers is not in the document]
woutslakhorst: I'll just remove it since it doesn't add anything.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,150 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+The message MUST contain a `token` that is to be used in the response.
+
+If the deconstruction fails, the local node sends a new `AdvertState` message. The `LC` value MUST be `512*floor((LC_req-1)/512)` where `floor` is a function that rounds a double down to the nearest integer. 
+The `XOR` value MUST be calculated over the transaction references in the range starting with 0 and ending with the new `LC` value.
```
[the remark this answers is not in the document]
woutslakhorst: Change point 2 in 4.2:
`* When the XOR is the same and the LC value equals BOB's highest Lamport Clock value: no action required.`

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+For a small network, 1024 buckets and 6 hashes is quite heavy. 
+For a nation-wide network this would be reasonable.
+These parameters are chosen to reduce the chance of a collision: two different transactions that are hashed into the exact same buckets.
+If there is a collision in the _set difference_, it is not be possible to deconstruct an IBLT.
+In the protocol described here, this would mean that an entire range of transactions would have to be sent.
+When this happens randomly, it wouldn't be such a problem, but it's also an attack vector.
+An attacker could craft pairs of transactions that would trigger other nodes into requesting large amounts of transactions. 
+This type of attack is called the [birthday attack](https://en.wikipedia.org/wiki/Birthday_attack).
+The only way to protect against this type of attack is to make the set large enough that the cost to the attacker would become to big.
```
reinkrul, line 202: too high

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+For an IBLT with 64 buckets and 4 hashes, the change for a collision is 50% after adding ~5000 keys.
```
reinkrul, line 203: change -> chance

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+This would be trivial for current computers.
+Expanding this to 1024 buckets and 6 hashes, reaching a 50% chance for a collision is only achieved after adding more than a billion keys.
+This is still not much, but the keys used for the collision have to be created by real signed transactions.
+Even on modern hardware, this could take an hour.
```
reinkrul, line 207: Are we quantum-proof? :-p

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+See Appendix A.1 for an explanation.
+
+### 4.1.3 IBLT page size
+
+Calculating an IBLT is relatively cheap, but it grows linearly with the number of transactions. This would harm performance of larger networks.
+The immutable nature of transactions make it possible to store intermediate IBLTs and use those as an optimization.
+The LC range parameters of the protocol should be adjusted accordingly to optimize use of these intermediate IBLTs.
+Therefore, they should not be chosen freely.
+
+Any LC value used in the protocol MUST be a multiple of 512 (minus 1, because the LC is 0 based).
```
reinkrul, line 102: This reads like a TXs LC can only be 0, 521, 1042, etc. BUt I'm sure that's not what you mean?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+1. Alice broadcasts at a set interval (§4.2.1):
+    * XOR of all known transaction references.
+    * Highest Lamport Clock value over all transactions (LC).
+    
+2. When receiving Alice's broadcast, Bob compares the XOR value from Alice with its own (§4.2.2):
+    * When the XOR is the same and the LC value equals BOB's highest Lamport Clock value: no action required.
+    * When different, send a message containing:
+      * LC value sent by Alice
+      * the IBLT that includes the Lamport Clock range of 0-LC
```
reinkrul, line 121: in other words, Bob sends his IBLT spanning his DAG from `LC=0` to `LC=LC(Alice)`?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+      * Bob's highest Lamport Clock value
+
+3. When receiving the message, Alice subtracts Bob's IBLT from her own IBLT for the given range (§4.2.3):
+   * If not deconstructable, go to 1 and send values based on a lower Lamport Clock value.
```
reinkrul, line 125: "a lower value" is vague for a spec. TODO?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -26,17 +28,24 @@ This document is released under the [Attribution-ShareAlike 4.0 International \(
 
 
 ## 2. Terminology
-
-* **peer**: remote Nuts node that communicates with the local node.
-* **node identity**: (a.k.a. node DID) the DID a node uses to identify itself on the Nuts network. Multiple logical nodes (e.g. a cluster) may share the same node identity.
-* **private transactions**: transactions that are meant for a specific receiver (node) on the network. The transaction payload is solely shared with that particular node.
+ 
+* **DAG** \(Directed Acyclic Graph\): graph formed of all transactions. It provides casual ordering for transactions and means to efficiently compare the local DAG with those of peers. See [RFC004](rfc004-verifiable-transactional-graph.md) for details.
+* **Invertible Bloom Lookup Table (IBLT)**: A bloom filter where instead of just a 0 or 1, a count, key_sum and hash_sum is stored per bucket. Original paper: https://arxiv.org/pdf/1101.2245 or https://www.ics.uci.edu/~eppstein/pubs/EppGooUye-SIGCOMM-11.pdf.
```
reinkrul, line 33: makes these proper links

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+The IBLT is a data structure capable of finding differences between sets.
+The performance of this data structure is not impacted by the number of entries, but only by the size of the set difference.
+This allows two nodes to compare the entire set of transactions with a relative small data structure.
+
+An IBLT has several parameters that define its characteristics. When comparing two IBLTs, these parameters MUST be the same.
+An overview of the parameters:
+
+* **buckets**: defines the size of the IBLT, much like the size of a bloom filter. The larger the size, the bigger the set difference can be. 
+* **Hk**: each key is hashed to select the buckets the key is inserted to. The hash function is applied on the previous output to create extra hash values. This continues until k different values are found. The hash function, its seed and the value for k MUST all be the same for the entire network.
+* **Hc**: each key is hashed to a checksum which is used to validate the inserted key during the deconstruction phase.
```
reinkrul, line 82: "deconstruction" is not defined (no idea how it works/what it is), not by the linked papers

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+    * When the XOR is the same and the LC value equals BOB's highest Lamport Clock value: no action required.
+    * When different, send a message containing:
+      * LC value sent by Alice
+      * the IBLT that includes the Lamport Clock range of 0-LC
+      * Bob's highest Lamport Clock value
+
+3. When receiving the message, Alice subtracts Bob's IBLT from her own IBLT for the given range (§4.2.3):
+   * If not deconstructable, go to 1 and send values based on a lower Lamport Clock value.
+   * If deconstructable, send a request for missing transactions
+   * If Bob's highest Lamport Clock value is higher that the LC value sent in 1, then request transactions by Lamport Clock value:
```
reinkrul, line 127: that -> than

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+     * Lamport Clock start value.
+     * Lamport Clock end value.
+
+4. When receiving a request for transactions, Bob responds with a message including the requested transactions (§4.2.4 and §4.2.5).
+
+5. After adding Bob's transactions to the DAG \(making sure its cryptographic signature is valid\), query the payload if it's missing.
```
reinkrul, line 133: not just signatures, but all validations. Also we're going atomic, aren't we?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+A node MUST make sure to only add transactions of which all previous transactions are present.
```
reinkrul, line 135: Does the protocol above assure that transactions are sync'd in order?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+The sections below specify the details of the protocol operation and maps it to the gRPC messages. See the section "Protobuf Definition" for a full specification of the protobuf/gRPC contract.
+
+### 4.2.1 Broadcasting
+
+The local node's DAG state MUST be broadcast at an interval using the `AdvertState` message, by default every 2 seconds. The interval MAY be adjusted by the node operator but MUST conform to the limits \(min/max interval\) defined by the network. It is advised to keep it relatively short because it directly influences the speed by which new transactions are retrieved.
```
reinkrul, line 141: I'd call the message `State`. Or something else, at least without the "Advert" part, since it describes the action (adverting), which just means it is a broadcast to all peers. So irrelevant for the message name, and it might also be triggered by the request from the peer for a specific range?

reinkrul, line 141: every 2 seconds, given 40k bytes = 40000/1024*30/1024 = 1,44mb per minute. Sounds OK-ish?

Although, that's for every peer. So for 10 peers that's 14,4mb. Hopefully our PIs don't blow up.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+The `AdvertState` message contains a `XOR` value, an `LC` value and a `token`. The `token` is scoped to a connection and is required in the response from the peer.
```
reinkrul, line 143: We could introduce `token` as v2 message header, applied to all query-response messages. Sounds like a "conversation ID"?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+The `LC` value MUST equals the highest Lamport Clock value of all transaction references included in the `XOR` calculation.
+If no transactions are present, an all-zero `XOR` and `LC` of 0 is sent.
+A node is free to choose how the `token` is created as long as it's able to satisfy the requirements of §4.2.3.
+It MUST be valid for at least 10 seconds.
```
reinkrul, line 147: what if it is received and it doesn't match?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+### 4.2.2 IBLT response
+
+When a peer receives an `AdvertState` message and the `XOR` differs, it SHOULD respond with a `TransactionSet` message.
+The upper bound of the response IBLT MUST be the equal to `PSIZE*ceil(LC/PSIZE)` where `LC` is the value from the `AdvertState` message and `ceil` is a function that rounds a double up to the nearest integer.
```
reinkrul, line 152: what are bounds of the IBLT?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+If `LC` equals zero, `PSIZE` is used.
+This upper bound is exclusive, the lower bound is inclusive and always 0.
```
reinkrul, line 154: 0, meaning from DAG root?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+Next to the IBLT data, the `TransactionSet` message MUST contain the original `LC` value as `LC_req`, a new `LC` value indicating the highest LC value from the peer and the `token` value from the `AdvertState` message.
```
reinkrul, line 155: since we're using protobuf, there's not really a need to shorten field names imo

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+For every `TransactionSet` message a node receives, it MUST check the `token` value in the message and compare it to its own administration.
+If the `token` doesn't exist or if the `LC_req` value doesn't match, the node MUST ignore the message.
+
+The IBLT in the `TransactionSet` message contains every transaction of the peer in the LC range of `0-LC_req`. The node MUST lookup/compute the IBLT for this range and subtract it from the IBLT of the peer. Deconstruction of the resulting IBLT will list the transaction refs the peer has and the local node misses. 
+
+These transactions MUST be queried by using the `TransactionListQuery` message. It contains all missing transaction references.
+The message MUST contain a `token` that is to be used in the response.
+
+If the deconstruction fails, the local node sends a new `AdvertState` message. The `LC` value MUST be `PSIZE*floor((LC_req-1)/PSIZE)` where `floor` is a function that rounds a double down to the nearest integer. 
+The `XOR` value MUST be calculated over the transaction references in the range starting with 0 and ending with the new `LC` value.
```
reinkrul, line 168: wouldn't it be pretty expensive to recalculate the XOR over a billion TXs?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+### 4.2.4 Transaction Range Query
+
+If the IBLT from a `TransactionSet` message could be deconstructed, the local node MAY also send a `TransactionRangeQuery` message.
+This is only needed if the `LC` value of the `TransactionSet` message is equal or larger than `PSIZE*ceil(LC_req/PSIZE)`. 
+This means the peer has additional transactions outside the IBLT range.
+The `TransactionRangeQuery` message contains a `begin` and `end` parameter indicating a request for transactions which LC value is between those parameters.
```
reinkrul, line 175: earlier we had `start` and `end` (contraty to `begin`)

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+The `begin` MUST be set to `PSIZE*ceil(LC_req/PSIZE)` and the end MUST be set to `LC`. Both are inclusive.
```
reinkrul, line 176: kinda confusing that `end` is inclusive here, and non-inclusive in the earlier IBLT range?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+Next to the IBLT data, the `TransactionSet` message MUST contain the original `LC` value as `LC_req`, a new `LC` value indicating the highest LC value from the peer and the `token` value from the `AdvertState` message.
+
+### 4.2.3 Transaction List Query
+
+For every `TransactionSet` message a node receives, it MUST check the `token` value in the message and compare it to its own administration.
+If the `token` doesn't exist or if the `LC_req` value doesn't match, the node MUST ignore the message.
+
+The IBLT in the `TransactionSet` message contains every transaction of the peer in the LC range of `0-LC_req`. The node MUST lookup/compute the IBLT for this range and subtract it from the IBLT of the peer. Deconstruction of the resulting IBLT will list the transaction refs the peer has and the local node misses. 
+
+These transactions MUST be queried by using the `TransactionListQuery` message. It contains all missing transaction references.
```
reinkrul, line 164: if we don't allow adding TXs with missing prevs to the DAG, you'll almost never query "random" TXs (only in case of a branch), so wouldn't range queries suffice?

reinkrul (changes requested):
I don't feel like I could implement (yet) it after reading this. It's complex stuff, but at this moment I feel like having to read both whitepapers to implement the protocol. Not sure to what end we should accomodate the reader here?

woutslakhorst:
> I don't feel like I could implement (yet) it after reading this. It's complex stuff, but at this moment I feel like having to read both whitepapers to implement the protocol. Not sure to what end we should accommodate the reader here?

We already refer to lots of specs and tech and assume that is familiar to the reader. If you're not familiar with gRPC, you'll have to read the gRPC manual and get familiar with the tools to generate code. It's also the same for stuff like ECIES. It's all a matter of level of knowledge.

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: If someone creates a branch at LC=0 while your LC is at 391867540913, then you have a big problem if your only tool is the range query. It could also have created a TX with prevs=[0, 391867540913]. With the IBLT you can exactly query those 2 TXs, with ranges not so much....

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -26,17 +28,24 @@ This document is released under the [Attribution-ShareAlike 4.0 International \(
...
 
 ## 2. Terminology
-
-* **peer**: remote Nuts node that communicates with the local node.
-* **node identity**: (a.k.a. node DID) the DID a node uses to identify itself on the Nuts network. Multiple logical nodes (e.g. a cluster) may share the same node identity.
-* **private transactions**: transactions that are meant for a specific receiver (node) on the network. The transaction payload is solely shared with that particular node.
+ 
+* **DAG** \(Directed Acyclic Graph\): graph formed of all transactions. It provides casual ordering for transactions and means to efficiently compare the local DAG with those of peers. See [RFC004](rfc004-verifiable-transactional-graph.md) for details.
+* **Invertible Bloom Lookup Table (IBLT)**: A bloom filter where instead of just a 0 or 1, a count, key_sum and hash_sum is stored per bucket. Original paper: https://arxiv.org/pdf/1101.2245 or https://www.ics.uci.edu/~eppstein/pubs/EppGooUye-SIGCOMM-11.pdf.
+* **Lamport Clock (LC)**: Monotonically increasing clock as defined in [RFC016](rfc016-lamport-clock.md).
```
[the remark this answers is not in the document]
woutslakhorst: `logical clock` then

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
 
-## X. Private Transactions
\ No newline at end of file
+## 4. Set reconciliation protocol
+
+### 4.1 Data requirements
+
+The protocol requires an XOR value and IBLT structure to be sent in a message. 
+These are to be calculated over different transaction ranges based on LC value.
+Both data structures have associative properties allowing for optimized data storage.
```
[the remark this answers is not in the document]
woutslakhorst: I'll remove it and let it up to the reader to apply a binary tree type of storage ;)

rfc/rfc017-distributed-network-grpc-v2.md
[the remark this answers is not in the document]
gerardsn: Yes, it should be uniform on random inputs. I just don't know if the hash-RNG provides any guarantees about how many numbers are generated before it repeats itself. This period could provide less than the k buckets needed to put the key/TxRef in the IBLT.

rfc/rfc017-distributed-network-grpc-v2.md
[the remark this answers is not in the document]
gerardsn: I find it easier to read it like this then having to find the details in the text. But that is a personal preference, so adding this provides options for different types of readers.

rfc/rfc017-distributed-network-grpc-v2.md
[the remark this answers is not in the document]
gerardsn: `LC` in `AdvertState` can be `LC_req` as well. 
If the IBLT cannot be decoded, a new `AdvertState` with lower `LC` will be send. This is not a _advert_ of the nodes LC, but a request for which LC range the peer should send an IBLT.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+The IBLT is a data structure capable of finding differences between sets.
+The performance of this data structure is not impacted by the number of entries, but only by the size of the set difference.
+This allows two nodes to compare the entire set of transactions with a relative small data structure.
+
+An IBLT has several parameters that define its characteristics. When comparing two IBLTs, these parameters MUST be the same.
+An overview of the parameters:
+
+* **buckets**: defines the size of the IBLT, much like the size of a bloom filter. The larger the size, the bigger the set difference can be. 
+* **Hk**: each key is hashed to select the buckets the key is inserted to. The hash function is applied on the previous output to create extra hash values. This continues until k different values are found. The hash function, its seed and the value for k MUST all be the same for the entire network.
```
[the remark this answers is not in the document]
woutslakhorst: changed to `applied recursively`

rfc/rfc017-distributed-network-grpc-v2.md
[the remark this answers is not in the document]
woutslakhorst: copied 2 of the 3 requirements for Hk

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,154 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+* **Hc**: each key is hashed to a checksum which is used to validate the inserted key during the deconstruction phase.
+* **data types**: the data types for *count*, *val_sum* and *hash_sum* MUST match ((un)signed, byte-size).
+
+The following parameters are used:
+
+* **buckets**: 1024
+* **Hc**: [murmur3](https://en.wikipedia.org/wiki/MurmurHash), 8 bytes, seed: 0x00
+* **Hk**: murmur3, 4 bytes, seed: 0x01, k = 6 
+* **key sum size**: 32 bytes
```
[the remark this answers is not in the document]
woutslakhorst: `val_sum` field size and `count` field size. Under `data_types`

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: The idea is to make the cost high enough and the impact low enough. The network will also be able to sync those transactions in a second. The cost of using a quantum computer will be high enough for the next 10 years or so?

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: The LC value in the message, not on the TX

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: introduced the concept of a page in the spec (not in the message), so we can use the notion of a "previous page"

rfc/rfc017-distributed-network-grpc-v2.md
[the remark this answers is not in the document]
gerardsn: If a node gets `infected` with the collision pair, it cannot reconcile new transactions from a peer that has not been infected. This slows down propagation of new Txs throughout the network until everyone is infected.

The current protocol only describes a way to detect that the nodes are not in sync, followed by set-reconciliation. I propose we broadcast TxRefs of recently received Txs also. This could be as a separate message, or add a []TxRef to the `AdvertState`. The peer can check if it is missing any of these and request them with a `TransactionListQuery`. This should resolve the differences in most cases and reduce the need for 'heavier' set-reconciliation.

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: It's called decoding there. Changed it to that.

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: That has been intentionally left out of this PR

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: That requirement has been added to 4.2.5 Transaction List message

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: Talked it over with @gerardsn , will add a Gossip message as well. This state message can then be used less often and only when the network is calm.

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: It is, could be part of another PR when another message type requires it?

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: stated in 4.2.3
when upgraded to conversation ID, it'll be in its own chapter.

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: refrased it

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: rephrased it using pages.

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: the implementation should store intermediate results in a binary tree. The IBLTs will have a maximum overhead of 40k per 512 transactions.

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: fixed

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,192 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+The sent `LC` value falls within the bounds of a page.
+The response IBLT MUST be calculated over the transactions leading up to and including that page.
+Next to the IBLT data, the `TransactionSet` message MUST contain the original `LC` value as `LC_req`, a new `LC` value indicating the highest LC value from the peer and the `token` value from the `State` message.
+
+#### 4.2.3 Transaction List Query
+
+For every `TransactionSet` message a node receives, it MUST check the `token` value in the message and compare it to its own administration.
+If the `token` doesn't exist or if the `LC_req` value doesn't match, the node MUST ignore the message.
+
+The IBLT in the `TransactionSet` message contains every transaction of the peer in the LC range of `0-LC_req`. The node MUST lookup/compute the IBLT for this range and subtract it from the IBLT of the peer. Decoding the resulting IBLT will list the transaction refs the peer has and the local node misses.
```
gerardsn, line 175: The IBLT received covers the range `0-min(LC, LC_req)`
if `LC < LC_req` we only need to compare to our own IBLT with range `0-LC`. If the LCs are far apart it will not be possible to decode the subtracted IBLTs.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,192 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+For every `TransactionSet` message a node receives, it MUST check the `token` value in the message and compare it to its own administration.
+If the `token` doesn't exist or if the `LC_req` value doesn't match, the node MUST ignore the message.
+
+The IBLT in the `TransactionSet` message contains every transaction of the peer in the LC range of `0-LC_req`. The node MUST lookup/compute the IBLT for this range and subtract it from the IBLT of the peer. Decoding the resulting IBLT will list the transaction refs the peer has and the local node misses. 
+
+These transactions MUST be queried by using the `TransactionListQuery` message. It contains all missing transaction references.
+The message MUST contain a `token` that is to be used in the response.
+
+If the decoding fails, the local node sends a new `State` message. The `LC` value MUST be the `end` of the page before the page that included `LC_req`.
```
gerardsn, line 180: The new `LC` in `State` can be one page lower than the page containing min(`LC`, `LC_req`) from the `TransactionSet`

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,192 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+      * Bob's highest Lamport Clock value
+
+3. When receiving the message, Alice subtracts Bob's IBLT from her own IBLT for the given range (§4.2.3):
+   * If not decodable, go to 1 and send values based on the previous page.
+   * If decodable, send a request for missing transactions
+   * If Bob's highest Lamport Clock value is higher than the LC value sent in 1, then request transactions by Lamport Clock value:
+     * Lamport Clock start value.
+     * Lamport Clock end value.
+
+4. When receiving a request for transactions, Bob responds with a message including the requested transactions (§4.2.4 and §4.2.5).
```
gerardsn, line 144: shouldn't §4.2.4 map to the last bullet of point 3?

gerardsn (commented):
With the addition of a gossip protocol for fast propagation of new Txs my main concern is addressed.

Just some minor things left

rfc/rfc017-distributed-network-grpc-v2.md
reinkrul: Moving it would break the v1 version of the protocol though (unless you keep it at both message and envelope level)

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -199,9 +220,26 @@ In the protocol described here, this would mean that an entire range of transact
...
 This is still not much, but the keys used for the collision have to be created by real signed transactions.
 Even on modern hardware, this could take an hour.
+
+### A.2 The right message for the right situation
+
+The protocol uses a variety of message to synchronize all transactions over nodes.
+Together they cover all known situations.
+
+If all nodes are operating correctly, the gossip protocol is the most efficient in synchronizing all the transactions.
+It will not be able to deal with situations where a node is offline or when nodes can no longer reach each other.
```
reinkrul, line 314: where a node _was_ offline (or has been)

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -199,9 +220,26 @@ In the protocol described here, this would mean that an entire range of transact
...
+It can be used frequent and uses small messages.
```
reinkrul, line 315: frequently because it uses small messages

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -199,9 +220,26 @@ In the protocol described here, this would mean that an entire range of transact
...
+
+If a node has been offline for some time, the combination of an IBLT and a range query will synchronize the node efficiently.
+The IBLT will resolve all missing transactions for the node's latest page and the range query will sycnhronize all the missing pages. Those pages will include all transactions that have been created during the period the nodes was offline.
```
reinkrul: sycnhronize  -> synchronize

reinkrul: nodes -> node

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -199,9 +220,26 @@ In the protocol described here, this would mean that an entire range of transact
...
+
+If a node has not been offline, but its communication has been interrupted, it will have created new transactions.
```
reinkrul, line 320: _might_ have created new transactions that aren't synchronized with the rest of the network

reinkrul, line 320: although one could argue that "communication interruption" is a form of being offline (its all about perspective)

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -199,9 +220,26 @@ In the protocol described here, this would mean that an entire range of transact
...
+The Lamport Clock values will overlap with the transactions of the rest of the network.
+The IBLT will be able to determine missing transactions. 
+If the difference in transactions is to big, the protocol will reduce the range the IBLT covers until an IBLT has been found that can resolve the transaction set difference.
```
reinkrul, line 323: is too great

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -54,26 +54,116 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+A node's new transaction references MUST be broadcast at an interval using the `Gossip` message, by default every 2 seconds. The interval MAY be adjusted by the node operator but MUST conform to the limits (min/max interval) defined by the network. It is advised to keep it relatively short because it directly influences the speed by which new transactions are retrieved.
+
+The `Gossip` message MUST contain an `XOR` value, an `LC` value and a list of transaction references (`transactions`).
+The list MUST contain all new transaction references since the last `Gossip` message.
+The list of transaction references MUST be tracked per connection.
+If no new transactions have been added/received or for the first message for a connection, an empty list is sent.
+The list MUST not contain more than 100 transactions.
+This could create a backlog of messages at the sending node's side. 
+A node SHOULD take precautions on keeping this backlog to a minimum, eg it SHOULD reduce the inflow of new transactions.
```
reinkrul, line 118: If the purpose of reducing the inflow is to keep the gossip protocol working, I'd say the functioning of the gossip protocol is less important than the inflow of transactions. So let the transactions come, but either skip some for the gossip protocol (let it be handled by the IBLT) or just keep sending them while respecting the limits of the network. You'll get 100 TXs every 2 seconds, so 3000 per minute. That's a lot of new TXs.

Yes, the creation of new TXs should be limited/throttled and the TX format should allow for batches for those cases its justified (e.g. new vendor publishing many, many private VCs).

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -54,26 +54,116 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+#### 5.2.2 Transaction List Query
+
+Upon receiving a `Gossip` message from a peer, a node MUST first compare the `XOR` value from the message with its own `XOR` value from the DAG.
+If the values are equal, no further action is required.
+
+If they are not equal, the transaction list MUST be filtered. All known transaction references MUST be removed.
+The resulting list contains all transaction the node is missing.
+The node MUST send a `TransactionListQuery` message containing the list of missing transaction references.
+This message MUST also add a `conversationID`.
```
reinkrul, line 133: The `Gossip` message lacks the `conversationID`, so you can't add it here?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -54,26 +54,116 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+## 4. Conversations
 
-One of the problems in a distributed network is how to make sure every node has processed all transactions.
-When two nodes haven't processed the same set of transactions, the second problem is how to efficiently synchronize the missing transactions.
-The higher the efficiency, the more transactions the entire network can process.
-The protocol described here is a compromise on efficiency and fairness. It's able to quickly determine a set of missing transactions, but also doesn't favour any node.
+Certain messages in the protocol are a response to a certain request. To make sure a response corresponds to a certain request, a `conversationID` is added to the request. Response type messages MUST add the same `conversationID` in the response.
+The `conversationID` is scoped to a connection.
+A node is free to choose the form of a `conversationID`.
+It MUST be valid for at least 10 seconds and no longer than 30 seconds.
```
reinkrul, line 62: is this long enough for longer IBLT interactions?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -54,26 +54,116 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
-## 4. Set reconciliation protocol
+## 4. Conversations
 
-One of the problems in a distributed network is how to make sure every node has processed all transactions.
-When two nodes haven't processed the same set of transactions, the second problem is how to efficiently synchronize the missing transactions.
-The higher the efficiency, the more transactions the entire network can process.
-The protocol described here is a compromise on efficiency and fairness. It's able to quickly determine a set of missing transactions, but also doesn't favour any node.
+Certain messages in the protocol are a response to a certain request. To make sure a response corresponds to a certain request, a `conversationID` is added to the request. Response type messages MUST add the same `conversationID` in the response.
+The `conversationID` is scoped to a connection.
+A node is free to choose the form of a `conversationID`.
```
reinkrul, line 61: I'd say it MUST be unique? Or unique enough, so that for at least the validity period there are no duplicates?

To avoid implementations using a hardcoded string (or counting 0-9 and then rolling over).

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -54,26 +54,116 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+It MUST be valid for at least 10 seconds and no longer than 30 seconds.
 
+If a node receives a response with a `conversationID`, it MUST match its contents with the original request.
```
reinkrul, line 64: Conversation IDs are bound to a specific connection/peer?

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -54,26 +54,116 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+The node MUST send a `TransactionListQuery` message containing the list of missing transaction references.
+This message MUST also add a `conversationID`.
+
+If the resulting list of transactions is empty and the `LC` value in the message is equal to or higher than the highest Lamport Clock value of the node, then the node MUST send a `State` message. See §6.2.1.
+
+#### 5.2.3 Transaction List
+
+When a node receives a `TransactionListQuery` message, it MUST respond with a `TransactionList` message.
+This is a response type message so it MUST include the sent `conversationID`.
+Each transaction in the message MUST have been requested. Transactions that resulted from a `TransactionListQuery` MUST have been present in that message.
```
reinkrul, line 141: stating the same thing twice, in other wording? Better write it once, but then in such a way it doesn't require a second wording.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -54,26 +54,116 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+#### 5.2.4 Transaction Payload query
+
+depends on private tx chaptr
+
+## 6. Set reconciliation protocol
+
+One of the problems in a distributed network is how to make sure every node has processed all transactions.
+When two nodes haven't processed the same set of transactions, the second problem is how to efficiently synchronize the missing transactions.
+The higher the efficiency, the more transactions the entire network can process.
+This part of the protocol is used to synchronize transactions that are missed by the gossip protocol.
```
reinkrul, line 158: (e.g. due to nodes being offline/network partitions)

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -54,26 +54,116 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+One of the problems in a distributed network is how to make sure every node has processed all transactions.
+When two nodes haven't processed the same set of transactions, the second problem is how to efficiently synchronize the missing transactions.
+The higher the efficiency, the more transactions the entire network can process.
+This part of the protocol is used to synchronize transactions that are missed by the gossip protocol. 
+
+### 6.1 Data requirements
+
+The protocol requires an XOR value and IBLT structure to be sent in a message.
+The XOR value has already ben explained in §5.1.1.
```
reinkrul, line 163: been

reinkrul (changes requested):

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: I'll rephrase it to keep the backlog below maximum. Because with more than 100TX per 2 seconds, it'll switch to a slower mechanism which will perform even less!

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: It's created here. A `TransactionListQuery` is the beginning of the conversation.

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: Each `State` is a new conversation.

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: as stated by line 60

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,270 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+4. After adding Alice's transactions to the DAG (making sure its cryptographic signature is valid), Bob SHOULD query the payload if it's missing.
+
+A node MUST make sure to only add transactions of which all previous transactions are present. If previous transactions are missing, Bob will send a `State` message.
+
+#### 5.2.1 Broadcasting new transactions
+
+A node's new transaction references MUST be broadcast at an interval using the `Gossip` message, by default every 2 seconds. The interval MAY be adjusted by the node operator but MUST conform to the limits (min/max interval) defined by the network. It is advised to keep it relatively short because it directly influences the speed by which new transactions are retrieved.
+
+The `Gossip` message MUST contain an `XOR` value, an `LC` value and a list of transaction references (`transactions`).
+The list MUST contain all new transaction references since the last `Gossip` message.
```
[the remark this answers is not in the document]
woutslakhorst: Could be interpreted as: your own new TXs not added to the DAG. I'll clarify

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,270 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+The list of transaction references MUST be tracked per connection.
+If no new transactions have been added/received or for the first message for a connection, an empty list is sent.
+The list MUST not contain more than 100 transactions.
+This could create a backlog of messages at the sending node's side. 
+A node SHOULD take precautions on keeping this backlog below maximum, eg it SHOULD reduce the inflow of new transactions.
```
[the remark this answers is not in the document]
woutslakhorst: That will have different consequences in different situations. You can't prefer your own TXs since they might depend on others previous transactions. It's also a SHOULD, not a MUST so it's up to the implementing party.

A restriction on inflow could be to pause requesting TXs from other nodes to give way to own TXs. If that is too much as well, a restriction on the API should occur. Also a buffer could be added in between allowing for some bursts of TXs. Then later when things cool down, messaging can catch up.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,270 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+If the values are equal, no further action is required.
+
+If they are not equal, the transaction list MUST be filtered. All known transaction references MUST be removed.
+The resulting list contains all transaction the node is missing.
+The node MUST send a `TransactionListQuery` message containing the list of missing transaction references.
+This message MUST also add a new `conversationID`.
+
+If the resulting list of transactions is empty and the `LC` value in the message is equal to or higher than the highest Lamport Clock value of the node, then the node MUST send a `State` message. See §6.2.1.
+
+#### 5.2.3 Transaction List
```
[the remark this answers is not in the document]
woutslakhorst: Last line of §5.2: 

```
A node MUST make sure to only add transactions of which all previous transactions are present. If previous transactions are missing, Bob will send a `State` message.
```

rfc/rfc017-distributed-network-grpc-v2.md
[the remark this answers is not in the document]
gerardsn: Restricting a node's intake of new Txs could cause the node to fall behind and potentially never catch up with the rest of the network. This could introduce a delay in authorization of access to patient data. I would prefer to reduce the interval for gossip messages in this situation.

The current requirements also forces all nodes, including 'spectators', to be an active participant in distribution of information. Generally speaking this is a good thing, but it might not be feasible for all to do so.

The exact gossip message probably requires some more thought. I prefer the message to have a minimal set of requirements so the exact form can be adapted when we learn more. I think the gossip messages should be more of the broadcast and forget type. It doesn't feel right that this has any impact on how/if the node accepts/creates new Txs.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,270 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+
+A node MUST make sure to only add transactions of which all previous transactions are present. If previous transactions are missing, Bob will send a `State` message.
+
+#### 5.2.1 Broadcasting new transactions
+
+A node's new transaction references MUST be broadcast at an interval using the `Gossip` message, by default every 2 seconds. The interval MAY be adjusted by the node operator but MUST conform to the limits (min/max interval) defined by the network. It is advised to keep it relatively short because it directly influences the speed by which new transactions are retrieved.
+
+The `Gossip` message MUST contain an `XOR` value, an `LC` value and a list of transaction references (`transactions`).
+The list MUST contain transaction references added to the DAG since the last `Gossip` message.
+The list of transaction references MUST be tracked per connection.
```
gerardsn, line 114: What is the purpose of this requirement? It sounds like an implementation detail.
```suggestion
```

rfc/rfc017-distributed-network-grpc-v2.md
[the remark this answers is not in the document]
gerardsn:
```suggestion
The list MUST not contain more than 100 transactions.
This could create a backlog of messages at the sending node's side. 
A node SHOULD take precautions to keep this backlog to a minimum.
```

gerardsn (approved):
I've added 2 more suggestions, but I can live without them.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,270 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+#### 5.2.3 Transaction List
+
+When a node receives a `TransactionListQuery` message, it MUST respond with a `TransactionList` message.
+This is a response type message so it MUST include the sent `conversationID`.
+Transactions that resulted from a `TransactionListQuery` MUST have been present in that message.
+Unknown transactions references SHOULD be ignored.
+Transactions that resulted from a `TransactionRangeQuery` MUST have an LC value that is within the requested range.
+If any of these requirements are not met, the entire message MUST be ignored.
+
+A `TransactionList` message MAY be broken up into smaller messages, each message should confirm to these rules. Each part MUST also use the same `conversationID`.
```
gerardsn, line 146:
```suggestion
A `TransactionList` message MAY be broken up into smaller messages, each message should conform to these rules. Each part MUST also use the same `conversationID`.
```

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +54,270 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
+The `XOR` value MUST be calculated over the transaction references in the range starting with 0 and ending with the new `LC` value.
+
+#### 6.2.4 Transaction Range Query
+
+If the IBLT from a `TransactionSet` message can be decoded, the local node MAY also send a `TransactionRangeQuery` message.
+This is only needed if the page containing the `LC` value of the `TransactionSet` message comes after the page containing `LC_req`. 
+This means the peer has additional transactions outside the IBLT range.
+If the `LC_req` value is in the latest page of the local node, it SHOULD query all pages leading up to the `LC` value.
+If the `LC_req` value is NOT in the latest page of the local node, then it MUST only query the next page.
+This last requirement prevents a node from querying the entire DAG while only some historic transactions are missing.
```
gerardsn, line 279: And solves the case where there is a collision on a page.

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: To make clear that you have to do this per connection. If all nodes do not do this, the gossip part will fail. We could change it to SHOULD to make it less strict.

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: added that too

PR #162, opened 2022-01-07
RFC017: Service discovery

reinkrul:
Questions to answer:

- [ ]  What if a node were to publish a competitor's `NutsComm` endpoint in its DID document, causing implementations to ignore the competitor's endpoint (because the malicious node's DID document is processed first). Maybe verify the `NutsComm` endpoint against the TLS certificate, so it can't be spoofed?
- [ ] Should we limit the number of (outbound) connections made?
- [ ] If we do, how do we prevent that all nodes connect to the first X nodes they discover, leaving other nodes undiscovered (random/nearest neighbour using hash distance).

But maybe, with the limited number of nodes we now have, we can safely let the nodes build a full mesh (or limit it to an arbitrarily chosen count, like 50)?

### DIFF
--- a/rfc/rfc017-distributed-network-grpc-v2.md
+++ b/rfc/rfc017-distributed-network-grpc-v2.md
@@ -469,6 +469,6 @@ If all nodes send at best-effort, that will be enough due to the number of nodes
 ## X. Service Discovery
 
 Although a (or multiple) trusted bootstrap node is required for initially connecting to the network, nodes SHOULD discover new peers by searching for `NutsComm` endpoints in the received DID documents.
-This SHOULD be done after the initial DAG sync is completed, to avoid connecting to non-existing endpoints resolved from older versions of DID documents.
-
-TODO: What if a node were to publish a competitor's NutsComm endpoint in its DID document, causing implementations to ignore the competitor's endpoint (because the malicious node's DID document is processed first)
+Only absolute `NutsComm` endpoints SHOULD be discovered. References to other DID documents, IP addresses, and domains as listed in [RFC2606](https://www.ietf.org/archive/id/draft-chapin-rfc2606bis-00.html) SHOULD be ignored.
+Connecting to discovered endpoints SHOULD be delayed until after the initial DAG sync is completed.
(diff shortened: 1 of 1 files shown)


woutslakhorst:
> What if a node were to publish a competitor's NutsComm endpoint in its DID document, causing implementations to ignore the competitor's endpoint (because the malicious node's DID document is processed first). Maybe verify the NutsComm endpoint against the TLS certificate, so it can't be spoofed?

The connection would error since the SAN from the certificate doesn't match => log and remove from wanted connection list.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -45,4 +45,11 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
...
 2. Assert that one of the `NutsComm` endpoint's host matches (one of) the `dNSName` SANs in the peer's TLS client certificate.
 
-## X. Private Transactions
\ No newline at end of file
+## X. Private Transactions
+
+## X. Service Discovery
+
+Although a (or multiple) trusted bootstrap node is required for initially connecting to the network, nodes SHOULD discover new peers by searching for `NutsComm` endpoints in the received DID documents.
+This SHOULD be done after the initial DAG sync is completed, to avoid connecting to non-existing endpoints resolved from older versions of DID documents.
```
woutslakhorst, line 53: NutsComm services with references to other DID Documents MUST be ignored. Only consider concrete endpoints.

woutslakhorst (changes requested):

woutslakhorst:
The other 2 points can be handled when a partial mesh is supported.

woutslakhorst:
@reinkrul status?

reinkrul:
> @reinkrul status?

Review vs. current implementation, update, review, merge

gerardsn:
@reinkrul Updated to describe service discovery as it is implemented

woutslakhorst (approved):

gerardsn:
Where should this go in the document? Probably §4 or §10, but I also see a case for generalizing the header of §3 and adding service discovery plus #254 there (if only because it is the smallest change to header numbers in a released version of this spec)

reinkrul:
Contents look OK, but propose includiong it in #255 and closing this PR

gerardsn:
superseded by #255

PR #168, opened 2022-02-14
Rename IRMA proof type name

dmeijboom:
~While working on the POC it felt like `NutsIrmaCredentialProof` wasn't the best name because `Nuts` is implied and `CredentialProof` is a little bit to generic. I'd like to rename it to `IRMASignatureProof` as that's what it is. It's a proof for IRMA credentials based on the IRMA signature.~

Edit: updated to `NutsIRMASignatureProof2022`

### DIFF
--- a/rfc/rfc013-vc-irma-proof.md
+++ b/rfc/rfc013-vc-irma-proof.md
@@ -1,10 +1,10 @@
 # RFC013 Verifiable Credential IRMA Proof Type
 
-|  |  |
-| :--- | :--- |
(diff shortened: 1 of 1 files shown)


dmeijboom:
After discussing the name we came up with `NutsIRMASignatureProof2022`

woutslakhorst (approved):

stevenvegt:
Should be `NutsIRMASignature2022` without the `Proof`

PR #174, opened 2022-04-12
RFC017: Specify TransactionRangeQuery response message

reinkrul:


rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -272,14 +272,15 @@ The `XOR` value MUST be calculated over the transaction references in the range
 #### 6.2.4 Transaction Range Query
 
 If the IBLT from a `TransactionSet` message can be decoded, the local node MAY also send a `TransactionRangeQuery` message.
+The peer MUST respond with `TransactionList` message, containing the requested transactions.
```
woutslakhorst, line 275: I'd say SHOULD. If the list is empty, nothing can be sent. If an error occurs, nothing is sent. The receiving party should be aware nothing can be sent.

thoughts?

gerardsn: We've had some discussions about only having one active (TransactionList) message at a time. If we decide to move in that direction, I think its better to have a way to mark a message done/complete without having to wait for the conversation to time out, as would be case when there is no response to a request.

woutslakhorst: An almost empty message does not mark a conversation as finished. Only if every requested TX is received (multi-part messages)

According to RFC2119:

SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

woutslakhorst: @reinkrul

PR #177, opened 2022-04-19
tie up loose ends for V2 protocol

### LINKED ISSUE #175: RFC017: XOR in state is confusing
https://github.com/nuts-foundation/nuts-specification/blob/9f633d07d245b16f077e2fbb7fd8835ccc46faae/rfc/rfc017-distributed-network-grpc-v2.md?plain=1#L270

The line above is confusing and conflicting because
- the description of `State` msg states that the XOR should be over the entire DAG.
- the peer always compares this XOR to that of its entire DAG, so the values under comparison do not (always) have the same meaning.

Proposal:
Change it so `State` always contains the XOR of the entire DAG, and the LC reflects the range to compare.  
This also creates a circuit breaker in case both DAGs get back in sync during set reconciliation.

woutslakhorst:
fixes #175

### DIFF
--- a/rfc/rfc017-distributed-network-grpc-v2.md
+++ b/rfc/rfc017-distributed-network-grpc-v2.md
@@ -90,18 +90,18 @@ The protocol generally operates as follows:
 1. Alice sends at a set interval (§5.2.1):
    * XOR of all known transaction references.
    * The highest LC value.
-   * A list of transaction references since the previous message. If no previous message has been send, the list is empty.
+   * A list of transaction references since the previous message. If no previous message has been sent, the list is empty.
 
 2. When receiving Alice's message, Bob compares the XOR value from Alice with its own (§5.2.2):
    * When the XOR is the same: no action required.
    * When different:
       * Remove all known transaction references from the list.
       * Request the remaining transactions.
-      * If the remaining list is empty and the LC value is equal or higher than Bob's highest LC value, Bob sends a `State` message as defined in Chapter 6. 
+      * Bob calculates a new XOR value combining its own XOR value and the newly requested transactions hashes. If this new XOR value differs and if Alice's LC is equal or higher thatBob's highest LC value, Bob sends a `State` message as defined in Chapter 6. 
 
 3. Alice responds with the requested transactions (§5.2.3).
 
-4. After adding Alice's transactions to the DAG (making sure its cryptographic signature is valid), Bob SHOULD query the payload if it's missing.
+4. After adding Alice's transactions to the DAG (making sure its cryptographic signature is valid), Bob SHOULD query the payload if it's missing. If Bob is missing transactions referenced by the received transactions, it sends a `State` message.
 
 A node MUST make sure to only add transactions of which all previous transactions are present. If previous transactions are missing, Bob will send a `State` message.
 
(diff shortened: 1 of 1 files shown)


rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -90,18 +90,18 @@ The protocol generally operates as follows:
...
       * Request the remaining transactions.
-      * If the remaining list is empty and the LC value is equal or higher than Bob's highest LC value, Bob sends a `State` message as defined in Chapter 6. 
+      * Bob calculates a new XOR value combining its own XOR value and the newly requested transactions hashes. If this new XOR value differs and if Alice's LC is equal or higher thatBob's highest LC value, Bob sends a `State` message as defined in Chapter 6. 
 
 3. Alice responds with the requested transactions (§5.2.3).
 
-4. After adding Alice's transactions to the DAG (making sure its cryptographic signature is valid), Bob SHOULD query the payload if it's missing.
+4. After adding Alice's transactions to the DAG (making sure its cryptographic signature is valid), Bob SHOULD query the payload if it's missing. If Bob is missing transactions referenced by the received transactions, it sends a `State` message.
 
 A node MUST make sure to only add transactions of which all previous transactions are present. If previous transactions are missing, Bob will send a `State` message.
```
gerardsn, line 106: The added line repeats the last part of line 106

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -241,15 +239,16 @@ A node MUST make sure to only add transactions of which all previous transaction
 
 The `State` message is sent as response to various conditions of the gossip protocol (see §5).
 The `State` message MUST contain a new `conversationID`.
-The `State` message contains a `XOR` value and an `LC`.
+The `State` message contains an `XOR` value and an `LC`.
 The `LC` value MUST equal the highest Lamport Clock value of all transaction references included in the `XOR` calculation.
 If no transactions are present, an all-zero `XOR` and `LC` of 0 is sent.
 
-This might look exactly like the `Gossip` message, but the type of response is different so it is a different type of message.
+This might look exactly like the `Gossip` message, but the type of response is different, so it is a different type of message.
```
gerardsn, line 246: Gossip also has a `transactions` field, so they don't look exactly the same to me

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -241,15 +239,16 @@ A node MUST make sure to only add transactions of which all previous transaction
...
 
 #### 6.2.2 IBLT response
 
 When a peer receives a `State` message and the `XOR` differs, it SHOULD respond with a `TransactionSet` message.
+This is a response type message, so it MUST include the sent `conversationID`.
```
gerardsn, line 251: this is the same as line 256. Verification of the conversationID (with one from `State`) when receiving a `TransactionSet` is missing in §6.2.3

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -241,15 +239,16 @@ A node MUST make sure to only add transactions of which all previous transaction
 
 The `State` message is sent as response to various conditions of the gossip protocol (see §5).
 The `State` message MUST contain a new `conversationID`.
-The `State` message contains a `XOR` value and an `LC`.
+The `State` message contains an `XOR` value and an `LC`.
```
gerardsn, line 242: To avoid confusion on what to send in response to a Gossip message, I think we should make clear that the XOR and LC reflect the current state of the local DAG, so not the temporary XOR calculated when parsing the Gossip msg.

gerardsn (changes requested):

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: Useless remark for a spec, will remove

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: added extra line for validation.

gerardsn (approved):

PR #182, opened 2022-04-21
Specify what errors may be returned in protocol v2

reinkrul:
Question is; what errors would we want to return at all? E.g.:

- TX was requested that the node doesn't have
- Peer sent empty TransactionListQuery
- No active conversation


woutslakhorst (approved):
Good to start with none and then see what needs arise to disclose some info

PR #189, opened 2022-05-06
RFC017: Extend TransactionList message and update flow

### LINKED ISSUE #185: RFC017: extend TransactionList message with MessageNumber and TotalMessages
All chunks of the `TransactionList` are received when `TransactionList.MessageNumber == TransactionList.TotalMessages`

This makes it possible to detect when a`TransactionRangeQuery` is done, simplifies the done detection for a `TransactionListQuery`, and solves some edge cases (peer can't sent requested Tx).

MessageNumber is 1 -based https://github.com/nuts-foundation/nuts-node/pull/1043

gerardsn:
fixes #185 
This also updates the message flow in the protocol so that only 1 active conversation is needed at a time.

### DIFF
--- a/rfc/rfc017-distributed-network-grpc-v2.md
+++ b/rfc/rfc017-distributed-network-grpc-v2.md
@@ -67,7 +67,8 @@ If a node receives a response with a `conversationID`, it MUST match its content
 If a `conversationID` is unknown or if the response doesn't match the requirements, the message MUST be ignored.
 The individual messages describe the requirements.
 
-A node SHOULD limit the number of conversations to a single node when these conversations contain overlapping data.
+A node SHOULD limit the number of conversations to a single peer when these conversations contain overlapping data.
+The protocol requires only 1 active conversation per peer at a time.
 It's up to the implementation on how to mark a conversation as finished and remove it from its administration.
 
 ## 5. Gossip protocol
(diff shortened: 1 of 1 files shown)


rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -101,8 +102,9 @@ The protocol generally operates as follows:
    * When the XOR is the same: no action required.
    * When different:
       * Remove all known transaction references from the list.
-      * Request the remaining transactions.
-      * Bob calculates a new XOR value combining its own XOR value and the newly requested transactions hashes. If this new XOR value differs and if Alice's LC is equal or higher thatBob's highest LC value, Bob sends a `State` message as defined in Chapter 6. 
+      * Bob calculates a new XOR value combining its own XOR value and the remaining new/unknown transactions hashes. 
+      * If this new XOR value matches that of Alice, or if Alice's LC is lower, Bob requests the list of unknown transactions.
```
woutslakhorst, line 106: This is independent of the LV value

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -293,15 +303,16 @@ The `XOR` value MUST still be calculated over all transaction references on the
 
 #### 6.2.4 Transaction Range Query
 
-If the IBLT from a `TransactionSet` message can be decoded, the local node MAY also send a `TransactionRangeQuery` message.
+If the IBLT from a `TransactionSet` message can be decoded but contains no missing transactions, the local node SHOULD send a `TransactionRangeQuery` message.
```
woutslakhorst, line 306: because of the 1 conversation limit this is an OR?

woutslakhorst (changes requested):
1 error and 1 question

rfc/rfc017-distributed-network-grpc-v2.md
gerardsn: This is also part of the 1 conversation limit. The Gossip can only result in 1 new conversation, so a TxListQuery or a State message. 

If the new XORs match, all missing Tx should be requested to make the DAGs consistent.

If the new XORs still don't match, Bob should send a State message, but only if Alice is not behind. It doesn't make sense to also request missing Tx if that doesn't resolve their differences since these are likely to have missing prevs, and these missing Tx will be solved by in the State route as well. If Alice is behind, we could do nothing and require Alice to update everything first, or we can request the missing Txs from Alice we know Bob doesn't have. This describes the latter, which could also result in a State message if any of the requested Txs has missing prevs.

rfc/rfc017-distributed-network-grpc-v2.md
gerardsn: Yes, we only send a RangeQuery when there is no ListQuery

rfc/rfc017-distributed-network-grpc-v2.md
woutslakhorst: The XOR can never match when the LC value is lower, it can be ignored?

woutslakhorst (approved):

PR #197, opened 2022-07-18
RFC017: Specify gRPC status code when authentication fails

reinkrul:


woutslakhorst (commented):
Does this require code changes?
Why the `16`, didn't know we number our codes?

reinkrul:
> Does this require code changes? Why the `16`, didn't know we number our codes?

Yes, it requires a code change. The change is backwards compatible.

The 16 - Unauthenticated is a constant defined by gRPC. This allows us to be more precise in the errors we return (and easier assertions). We currently return an "unknown" error with a message specified by this RFC. I figured it would be nicer when we use a gRPC status code for it.

woutslakhorst (approved):

PR #198, opened 2022-07-28
changed requirement of TX to point to all prevs

woutslakhorst:
See appendix A.2 of rfc004 for rationale

### DIFF
--- a/rfc/rfc004-verifiable-transactional-graph.md
+++ b/rfc/rfc004-verifiable-transactional-graph.md
@@ -89,11 +89,11 @@ All transactions referred to by `prevs` MUST be present, since failing to do so 
 
 As the name implies the DAG MUST be acyclic, transactions that introduce a cycle are invalid MUST be ignored. ANY following transaction that refers to the invalid transaction \(direct or indirect\) MUST be ignored as well.
 
-Since it takes some time for the transactions to be synced to all network peers \(eventual consistency\) there COULD be multiple transactions referring to the previous transactions in the **prevs** field, a phenomenon called _branching_. Branching is common and harmless. Branches are merged by specifying their heads in the **prevs** field:
+Since it takes some time for the transactions to be synced to all network peers \(eventual consistency\) there COULD be multiple transactions referring to the previous transactions in the **prevs** field, a phenomenon called _branching_. Branching is common and relatively harmless. Branches are merged by specifying their heads in the **prevs** field:
 
 ![DAG structure](../.gitbook/assets/rfc004-branching.svg)
 
-Branches may remain unmerged.
+Branches may remain unmerged. As mentioned at the beginning of this paragraph, a transaction MUST refer to a previous transaction. It doesn't matter on which branch the latest transaction is published. 
 
 The first transaction in the DAG is the _root transaction_ and SHALL NOT have any **prevs** entries. There MUST only be one root transaction for a network and subsequent root transactions MUST be ignored.
 
--- a/rfc/rfc004-verifiable-transactional-graph.md
+++ b/rfc/rfc004-verifiable-transactional-graph.md
@@ -83,19 +83,21 @@ When serializing a reference to string form it MUST be hexadecimal encoded and S
 
 ### 3.3 Ordering, branching and merging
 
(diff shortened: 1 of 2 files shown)


rfc/rfc004-verifiable-transactional-graph.md
```diff
@@ -83,19 +83,21 @@ When serializing a reference to string form it MUST be hexadecimal encoded and S
...
 
 All transactions referred to by `prevs` MUST be present, since failing to do so would corrupt the DAG.
 
 As the name implies the DAG MUST be acyclic, transactions that introduce a cycle are invalid MUST be ignored. ANY following transaction that refers to the invalid transaction \(direct or indirect\) MUST be ignored as well.
 
-Since it takes some time for the transactions to be synced to all network peers \(eventual consistency\) there COULD be multiple transactions referring to the previous transactions in the **prevs** field, a phenomenon called _branching_. Since branches \(especially old and/or long ones\) may cause transactions to be reordered which hurts performance they MUST be merged as soon as possible. Branches are merged by specifying their heads in the **prevs** field:
+Since it takes some time for the transactions to be synced to all network peers \(eventual consistency\) there COULD be multiple transactions referring to the previous transactions in the **prevs** field, a phenomenon called _branching_. Branching is common and harmless. Branches are merged by specifying their heads in the **prevs** field:
```
reinkrul, line 92: I discussed with @gerardsn that having many parallel branches might hamper our set reconcilliation algorithm, due to the fact that you can have many, many TXs with the same lamport clock. So with our current protocol, it's not harmless?

reinkrul (changes requested):

rfc/rfc004-verifiable-transactional-graph.md
woutslakhorst: *relatively harmless*?

Branching will occur nevertheless. Merging them again doesn't have a positive effect on performance.

gerardsn: branching is fine as long as new Txs include the most recent LC + 1 as the clock value. If a random branch head is taken as prev to calculate the clock we will have additions to the DAG at any clock level. This against the assumptions made in the set reconciliation algorithm.

So a transaction MUST add the transaction with the highest LC value to its prevs. If there are multiple branches with the same LC, anyone of them is fine. I think this is already covered in these changes, but making this a bit more explicit would help

gerardsn (changes requested):

rfc/rfc004-verifiable-transactional-graph.md
gerardsn: ahh, and the other reason for referencing all prevs was to make sure the transactions used in by the new transaction are all present on the dag. We are already including the Tx containing the key used for signing the new Tx as a prev, are there more Txs containing data that must be present for a new Tx to be processed?

woutslakhorst: > are there more Txs containing data that must be present for a new Tx to be processed?
currently not

reinkrul (approved):

reinkrul:
Just thought of additional issues:

I see potential issues when nobody tries to merge unmerged heads:

- There's no incentive to limit the number of branches. While a single branch, or a bunch of them are harmless, 1000 branches with a TX with the same LC is not (due to page size). Now the number of branches can be every-increasing, while previously it would attempt to return to 1 branch
- There are dependencies between TXs not modelled in the prevs: VC subjects (maybe others as well?). A VC TX does not refer to the TX which created/last updated the subject's DID document, so it might be in a different branch. Given that TX are casually ordered, a TX (e.g. a VC) that refers to something that happened in another TX (a DID) must refer to that earlier TX, either directly (prev) or indirect (some 'prevs ago'). Previously this wasn't a problem, because you referred to all heads. That automatically included a potential branch which contains the dependent TX (e.g. the DID).

reinkrul (changes requested):
Found additional issue

gerardsn (approved):

woutslakhorst:
> There's no incentive to limit the number of branches. While a single branch, or a bunch of them are harmless, 1000 branches with a TX with the same LC is not (due to page size). Now the number of branches can be every-increasing, while previously it would attempt to return to 1 branch

When branches are merged, there's also no limit on the number of branches.

> A VC TX does not refer to the TX which created/last updated the subject's DID document, so it might be in a different branch

Yes it does: `network_publisher#88`
All dependencies on earlier transactions must be solved by prevs, ordering of transactions can never be used because that would be an attack vector.

reinkrul (approved):
Approved after discussion

PR #203, opened 2022-09-01
RFC004: add SHOULD requirement to prevs field

### LINKED ISSUE #201: RFC004: limit number of prevs
An implementation SHOULD limit the number of prevs. This can't be enforced yet..
The reference implementation already does this and limits it to the last + required prevs.

woutslakhorst:
closes #201 

rfc016 is merged into rfc004 to prevent chicken-and-egg-problem


rfc/rfc004-verifiable-transactional-graph.md
```diff
@@ -61,9 +61,10 @@ In addition to the registered header parameters, the following headers MUST be p
 
 * **sigt**: \(signing time\) MUST contain the signing time of the transaction as Unix time since epoch encoded as NumericValue.
 * **ver**: MUST contain the format version of the transaction as number. For this version of the format the version MUST be 1.
-* **prevs**: \(previous transactions\) MUST contain the references \(see section 3.2\) of the preceding transactions \(see section 3.4\).
+* **prevs**: \(previous transactions\) MUST contain references \(see section 3.2\) of preceding transactions \(see section 3.4\).
   * When it's a root transaction the field SHALL NOT have any entries.
   * When creating a transaction it MUST only contain transactions that the local node has successfully processed, to avoid publishing unprocessable transactions.
+  * The field SHOULD contain only one of the latest heads in addition to other requirements (for example requirements introduced by [RFC006 §5.1](rfc006-distributed-registry#51-conflict-detection)).
```
gerardsn, line 67: In RFC016 we state that the LC is calculated as 'prev with highest LC + 1'. The current formulation allows using any of the current heads. 

This might cause a conflict in LC calculation. If we only add 1 of the heads to the prevs, the list of heads will eventually grow out of control. 

* The field SHOULD contain at most 5 references 
* One of the references MUST be the head with the highest LC. If there are multiple heads with the same highest LC, at least one MUST be added as a reference.

One is going to be a reference to the Tx containing the key used for signing. That currently leaves 3 prevs that can be chosen freely and may be assigned to specific references later on.

woutslakhorst: it should not use the term 'head' but 'one of the transactions with the highest LC' There is no list of heads anymore so that won't be an issue.

woutslakhorst: It might be better indeed to place RFC016 content inside RFC004

reinkrul, line 67: maybe rephrase it to "SHOULD contain only one reference to a latest (with highest LC) transaction in addition to"?

rfc/rfc004-verifiable-transactional-graph.md
```diff
@@ -54,21 +55,28 @@ In addition to required header parameters as specified in RFC7515 the following
...
 In addition to the registered header parameters, the following headers MUST be present as protected headers:
 
 * **sigt**: \(signing time\) MUST contain the signing time of the transaction as Unix time since epoch encoded as NumericValue.
-* **ver**: MUST contain the format version of the transaction as number. For this version of the format the version MUST be 1.
-* **prevs**: \(previous transactions\) MUST contain the references \(see section 3.2\) of the preceding transactions \(see section 3.4\).
+* **ver**: MUST contain the format version of the transaction as number. For this version of the format the version MUST be 1 or 2.
+* **prevs**: \(previous transactions\) MUST contain references \(see section 3.2\) of preceding transactions \(see section 3.4\).
   * When it's a root transaction the field SHALL NOT have any entries.
   * When creating a transaction it MUST only contain transactions that the local node has successfully processed, to avoid publishing unprocessable transactions.
+* The **lc** MUST be a positive number constructed as follows:
```
gerardsn, line 69:
```suggestion
* **lc**: MUST be a positive number constructed as follows:
```

rfc/rfc004-verifiable-transactional-graph.md
```diff
@@ -54,21 +55,28 @@ In addition to required header parameters as specified in RFC7515 the following
 
 * **cty**: MUST contain the type of the transaction content indicating how to interpret it.
 * **crit** MUST contain the **sigt**, **ver**, **prevs** and **pal** headers.
```
gerardsn, line 57: Question, should pal always be present, or only when there is a pal header?

rfc/rfc004-verifiable-transactional-graph.md
```diff
@@ -54,21 +55,28 @@ In addition to required header parameters as specified in RFC7515 the following
...
+* **crit** MUST also contain the **lc** header if **ver** == 2.
```
gerardsn, line 58: Should this be:
```suggestion
* **crit** MUST also contain the **lc** header if **ver** >= 2.
```
A future RFC can remove this requirement.

rfc/rfc004-verifiable-transactional-graph.md
```diff
@@ -93,7 +101,8 @@ Since it takes some time for the transactions to be synced to all network peers
 
 ![DAG structure](../.gitbook/assets/rfc004-branching.svg)
 
-Branches may remain unmerged. As mentioned at the beginning of this paragraph, a transaction MUST refer to a previous transaction. It doesn't matter on which branch the latest transaction is published. 
+Branches may remain unmerged. As mentioned at the beginning of this paragraph, a transaction MUST refer to a previous transaction. It doesn't matter on which branch the latest transaction is published.
+**LC** values will overlap between branches.
```
gerardsn, line 105: I might be wrong, but it looks like we are only referencing 1 head. If that case the part about merging in section 3.3 is no longer relevant?

gerardsn (changes requested):

rfc/rfc004-verifiable-transactional-graph.md
gerardsn: to match the rest

rfc/rfc004-verifiable-transactional-graph.md
woutslakhorst: No, since there's no version >= 2

rfc/rfc004-verifiable-transactional-graph.md
woutslakhorst: yes

rfc/rfc004-verifiable-transactional-graph.md
woutslakhorst: It doesn't hurt either. There may be some content in there that might be relevant for garbage collection.

reinkrul (approved):

gerardsn (approved):

PR #206, opened 2022-09-05
update RFC006: conflict detection for create transactions

### LINKED ISSUE #115: RFC006 DID creation vulnerability
We have an attacker Eve which creates two DID Documents using the same keys. One DID Document has an added assertion key, the other has not. Eve will publish these DID Documents as two transactions. Both as create transactions with a JWK.

Other nodes will receive the transactions is some order.
We currently have no definition on when a TX is an update of a DID Document or a create. Neither do we have specified what should be done when two TXs create the same DID Document, but with different contents.

The current Nuts node will detect an update based on DID existence (not by checking jwk/kid header), meaning that the second DID Document received will not be updated in the VDR. When the assertion key is used to sign TXs, some nodes will not process the transaction since they do not have the assertion key, others will....

This is not a problem when the nuts-node is the only implementation. But it's an attack.

Solutions:
- define behaviour for parallel creates (with merges)
- use hashes to determine the DID (hash of DID Document without ID field). This also requires different identifiers for the keys and services. Prefixes of the DID are no longer possible for key material. Requires an extra DB search: KID to DID Document.

woutslakhorst:
fixes #115

### DIFF
--- a/rfc/rfc006-distributed-registry.md
+++ b/rfc/rfc006-distributed-registry.md
@@ -413,7 +413,7 @@ Since the information is self-proclaimed and not authenticated or verified in an
 
 ## 5.1 Conflict detection
 
-It's a conflict for a create transaction when:
+A transaction that creates a DID document causes a conflict when all the following conditions apply:
 
 * Two or more transactions exist that introduce the same DID (both have the same `jwk` header).
 * The contents of the DID Document differs.
--- a/rfc/rfc006-distributed-registry.md
+++ b/rfc/rfc006-distributed-registry.md
@@ -413,7 +413,12 @@ Since the information is self-proclaimed and not authenticated or verified in an
 
 ## 5.1 Conflict detection
 
-The algorithm for conflict detection can be summarized as follows:
+It's a conflict for a create transaction when:
+
+* Two or more transactions exist that introduce the same DID (both have the same `jwk` header).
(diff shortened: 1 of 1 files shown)


rfc/rfc006-distributed-registry.md
```diff
@@ -413,7 +413,12 @@ Since the information is self-proclaimed and not authenticated or verified in an
 
 ## 5.1 Conflict detection
 
-The algorithm for conflict detection can be summarized as follows:
+It's a conflict for a create transaction when:
```
reinkrul, line 416: I think both conditions must apply?
```suggestion
A transaction that creates a DID document causes a conflict when all the following conditions apply:
```

reinkrul (changes requested):

rfc/rfc006-distributed-registry.md
```diff
@@ -413,7 +413,12 @@ Since the information is self-proclaimed and not authenticated or verified in an
 
 ## 5.1 Conflict detection
 
-The algorithm for conflict detection can be summarized as follows:
+A transaction that creates a DID document causes a conflict when all the following conditions apply:
```
reinkrul, line 416:
```suggestion
A transaction that creates a DID document causes a conflict when all of the following conditions apply:
```

reinkrul (approved):
Minor suggestion

PR #208, opened 2022-09-05
moved NutsAuthorizationCredential specific validations to RFC014

### LINKED ISSUE #146: RFC003 validation for NutsAuthorizationCredential not totally clear.
The current text for 5.2.1.7:

```5.2.1.7 Validate verifiable credentials
The vcs field contains the list of verifiable credentials of any type.
If a credential has the NutsAuthorizationCredential type it determines whether the accessible resource and is valid according to the following rules (based on RFC014):
* The credential issuer equals the sub field of the JWT.
* The credential credentialSubject.id equals the iss field of the JWT.
* The credential proof is valid.
* The credential has not been revoked.
* The credential has not expired.
It MUST be verified that the requester's request conforms to the specified policy and only requests allowed resources.
```

The 2nd sentence has a typo.
The first point is the problematic one, because for other types of credentials this isn't true.
A case could be made that the credential issuer only needs to be trusted. The level of trust there should be at the DID level, not globally. This would enable a trusted third party to issue credentials on behalf of a single DID.

This is required for kik-v, babyconnect and ANW bolts

woutslakhorst:
closes #146

### DIFF
--- a/rfc/rfc014-authorization-credential.md
+++ b/rfc/rfc014-authorization-credential.md
@@ -139,7 +139,7 @@ These are in addition to the ones listed in [§5.2.1.7 of RFC003](rfc003-oauth2-
 * The credential `issuer` equals the **sub** field of the JWT in the access token request.
 * The credential `credentialSubject.id` equals the **iss** field of the JWT in the access token request.
 
-An NutsAuthorizationCredential is not needed when requester and authorizer are the same.
+A `NutsAuthorizationCredential` is not needed when `requester` and `authorizer` are the same.
 
 ### 4.3 Runtime enforcement
 
--- a/rfc/rfc014-authorization-credential.md
+++ b/rfc/rfc014-authorization-credential.md
@@ -119,9 +119,35 @@ Example:
 
 ## 4. Access Control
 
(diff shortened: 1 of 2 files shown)


rfc/rfc014-authorization-credential.md
```diff
@@ -119,9 +119,35 @@ Example:
...
+
+### 4.2 Access token request 
+
+The following validations are to be performed by the authorization server during an access token request.
+These are in addition to the ones listed in [§5.2.1.7 of RFC003](rfc003-oauth2-authorization.md#521-validation-steps).
+
+* The credential `issuer` equals the **sub** field of the JWT in the access token request.
+* The credential `credentialSubject.id` equals the **iss** field of the JWT in the access token request.
+
+An NutsAuthorizationCredential is not needed when requester and authorizer are the same.
```
reinkrul, line 142:
```suggestion
A `NutsAuthorizationCredential` is not needed when `requester` and `authorizer` are the same.
```

reinkrul (changes requested):

gerardsn (approved):

reinkrul (approved):

PR #231, opened 2022-11-03
RFC017: Readability and minor corrections

reinkrul:
- In chapter 5 (gossip), Alice gossips to Bob, which responds by starting set reconciliation. For readability, chapter 6 should keep using these roles to avoid confusion.
- 6.2: only compare XOR when receiving State message (according to implementation)

### DIFF
--- a/rfc/rfc017-distributed-network-grpc-v2.md
+++ b/rfc/rfc017-distributed-network-grpc-v2.md
@@ -265,7 +265,7 @@ The protocol operates by comparing transaction pages, starting at the last page 
 3. When receiving the message, Bob subtracts Alice's IBLT from his own IBLT for the given range (§6.2.3) and does one of the following:
    * If not decodable, go to step 1 and send State message for the previous page.
      * If already comparing the first page, request all transactions of the first page.
-   * If decodable, send a request for missing transactions if there are any.
+   * If decodable and missing transactions, send a request for missing transactions
    * If Alice's highest Lamport Clock value is higher than the LC value sent in 1, then request transactions over a range of Lamport Clock values (§6.2.4)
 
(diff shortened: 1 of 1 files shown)


rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -256,7 +256,7 @@ The protocol generally operates as follows:
     * Highest Lamport Clock value over all transactions (LC).
     
 2. When receiving Bob's message, Alice compares the XOR value from Bob with its own (§6.2.2):
-    * When the XOR is the same and the LC value equals Alice's highest Lamport Clock value: no action required.
```
gerardsn: This condition had a reason. Maybe @woutslakhorst can remember what that was?
(its not in the code (anymore))

gerardsn (changes requested):

rfc/rfc017-distributed-network-grpc-v2.md
reinkrul: Note that it is not implemented

woutslakhorst:
```// nothing to do if peers are now synced
	if xor.Equals(hash.FromSlice(msg.XOR)) {
		return nil
	}```
in `handlers.go#handleState`?

woutslakhorst: Might be a leftover when XOR was not over all transactions...

gerardsn: Makes sense. And would be fine to drop the requirement in that case

reinkrul:
Stale.

PR #255, opened 2023-03-06
RFC017: Add connections section

### LINKED ISSUE #252: RFC017 §3 disconnect on node authentication is incorrect
With dynamic discovery enabled, it won't be possible for a node to connect to a bootstrap node that changed its NodeDID somewhere in time.

The new node is synchronizing transactions and somewhere it gets the first (outdated) NutsComm address. Auto discovery will update and then try to reconnect/reauthenticate. This will fail, it should continue with connecting and fetch transactions.

Disconnect on failed authentication should be changed.

### LINKED ISSUE #254: RFC017: peerID is missing in spec
[§6.1. Peer Identification](https://nuts-foundation.gitbook.io/drafts/rfc/rfc005-distributed-network-using-grpc#6.1.-peer-identification) of RFC005 is missing in RFC017

gerardsn:
copied and adapted from RFC005 §6
closes #252
closes #254

### DIFF
--- a/rfc/rfc017-distributed-network-grpc-v2.md
+++ b/rfc/rfc017-distributed-network-grpc-v2.md
@@ -45,7 +45,15 @@ Connections are subject to the requirements specified by [RFC008 Certificate Str
 
 Other terminology comes from the [Nuts Start Architecture](rfc001-nuts-start-architecture.md#nuts-start-architecture).
 
-## 3. Authenticating node identity
+## 3. Connections
+
+The protocol uses [gRPC](https://grpc.io/) over [HTTP/2](https://tools.ietf.org/html/rfc7540). Since the protocol uses a bi-directional stream over which peers can send and receive messages, there needs to be only a single connection between two peers. For instance, if node _A_ connects to node _B_, _B_ can send messages back to _A_ without having to connect back to node _A_. Nodes SHOULD avoid maintaining duplicate connections to their peers to minimize traffic and system load.
+
+### 3.1. Peer Identification
+
+Since nodes might be operating behind reverse proxies, NAT routers and/or load balancers there needs to be a way to identify a peer that's not dependent on the remote host/port of the connection. Instead, a node MUST generate a globally unique identifier \(the node's _peer ID_\) and provide it as `peerID` in gRPC connection metadata on incoming or outgoing connections. A peer ID MAY be changed every time a node \(re\)starts but MUST be the same for all connections of that node.
+
+### 3.2. Authenticating node identity
 
 Particular exchanges (private transactions) might require authentication of the peer's identity. This identity (a.k.a. node identity) is specified by [RFC015 Node identity](rfc015-node-identity.md).
 If a node has a node DID and wishes to create an authenticated connection, it MUST send the DID as `nodeDID` gRPC header when establishing inbound or outbound connections.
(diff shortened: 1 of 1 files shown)


gerardsn:
I have left out RFC005's [6.4. Protocol Version](https://nuts-foundation.gitbook.io/drafts/rfc/rfc005-distributed-network-using-grpc#6.4.-protocol-version). We don't include the version number anymore, but I don't know if it has been replaced by something else.

woutslakhorst (approved):

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -57,6 +65,21 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
 1. Resolve peer's node DID to its corresponding DID document.
 2. Assert that one of the `NutsComm` endpoint's host matches (one of) the `dNSName` SANs in the peer's TLS client certificate.
 
+To prevent bootstrapping issues, a node MAY have an authenticated and an anonymous connection (both sides are unauthenticated) with a peer at the same time.
```
reinkrul, line 68: "sides" is ambiguous. It describes the perspective from 1 node, so if the remote node sees the local node as authenticated/unauthenticated is irrelevant for this sentence, I think.

rfc/rfc017-distributed-network-grpc-v2.md
```diff
@@ -57,6 +65,21 @@ As specified by RFC015, the node MUST authenticate the peer's node DID as follow
 1. Resolve peer's node DID to its corresponding DID document.
 2. Assert that one of the `NutsComm` endpoint's host matches (one of) the `dNSName` SANs in the peer's TLS client certificate.
 
+To prevent bootstrapping issues, a node MAY have an authenticated and an anonymous connection (both sides are unauthenticated) with a peer at the same time. 
+To create an anonymous connection, a peer MUST NOT send its `nodeDID` and MUST ignore any `nodeDID` returned by the peer.
```
reinkrul, line 69: Why the latter? If I never create a DID for my node (which is the case for my dev node), I can still have a connection on which I authenticated the remote node. We need clarification on connection auth status, which kinds do we have and what does it imply (e.g. both sides unauthenticated, one side authenticated, etc).

reinkrul (changes requested):

rfc/rfc017-distributed-network-grpc-v2.md
gerardsn: Before we accepted the bootstrap connection as authenticated even if authentication failed. This is undesirable for obvious reason, but solved an issue where the node does not have the data it needs to perform authentication (yet) and can't get the data because authentication fails. 

Making the bootstrap connection anonymous and allowing a second connection through service discovery solves the bootstrap issue without granting the bootstrap connection any extra privileges.

gerardsn: will rewrite to make this clearer

rfc/rfc017-distributed-network-grpc-v2.md
gerardsn: Nodes do not accept duplicate connections. To have an authenticated connection next to the bootstrap connection, both sides of the bootstrap connection need to be unauthenticated. If half of the connection is already authenticated, the node holding that connection is going to reject the second connection. This prevents a fully authenticated connection from being established -> no private VC exchange with bootstrap nodes.

gerardsn:
> I have left out RFC005's [6.4. Protocol Version](https://nuts-foundation.gitbook.io/drafts/rfc/rfc005-distributed-network-using-grpc#6.4.-protocol-version). We don't include the version number anymore, but I don't know if it has been replaced by something else.

@reinkrul any insights on this?

reinkrul:
> > I have left out RFC005's [6.4. Protocol Version](https://nuts-foundation.gitbook.io/drafts/rfc/rfc005-distributed-network-using-grpc#6.4.-protocol-version). We don't include the version number anymore, but I don't know if it has been replaced by something else.
> 
> @reinkrul any insights on this?

Protocol version is included in the protocol implementation (gRPC method name) itself, so no need to specify it in the message imo.

reinkrul (approved):

PR #256, opened 2023-04-04
RFC002 Add the NutsEmployeeIdentity means

stevenvegt:
This PR is a RFC for the `NutsEmployeeIdentity` which allows a care organisation to construct a claim about a current logged-in user. 
Since the user identity is issued by the care origanisation instead of a trusted third party (such as the government), it has a low trust level, but also comes with a simpler UX. It can be used in situations where a lower level of trust is appropriate and and then removes an extra burden from the care givers of using a personal authentication device.


woutslakhorst (commented):
RFC002 is part of spec v1.0. I think you need to create a new RFC that amends rfc002.
Also new sentences may be put on a new line, this will make it easier to spot differences in a diff when updated.

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,81 @@ Beside the mandatory VP fields, the following applies:
 * The `proof.type` field MUST equal `NutsUziSignedContract`.
 * The `proof.proofValue` field MUST contain the JWT in its compact serialization form as described in \[[RFC7515 section 3.1](https://tools.ietf.org/html/rfc7515#section-3.1)\].
 
+### 7.3 NutsEmployeeIdentity
+
+The NutsEmployeeIdentity method is an authentication _means_ which allows a care organisation to assert the identity of the current user. The care organisation uses the information of the current user session to present the user with a statement about its identity and asks the user to confirm this. After confirmation this bundle of information is singed with the care organisation's public key.
```
woutslakhorst, line 406: private key, not public key

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,81 @@ Beside the mandatory VP fields, the following applies:
...
+
+This method makes it possible to assert a user identity without the need for a personal authentication means. This puts the care organisation in the role of identity issuer instead of a trusted third party. A use-case (Bolt) can choose to allow this authentication under certain (low risk) circumstances which still require a person to be authenticated in order to comply with the requirements of the [NEN7513](https://www.nen.nl/nen-7513-2018-nl-245399)(Dutch standard for logging in context of medical ICT).
```
woutslakhorst, line 408: Just call it "applicable legislation"? So the rfcs remain relatively free of national context?

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,81 @@ Beside the mandatory VP fields, the following applies:
...
+
+Note: The _means_ is designed in a way that it is interchangeable with (other) means with a higher trust level, (using a personal authentication device). Even it is technically possible, this means MUST NOT be used to perform machine to machine interactions which are not initiated by the current user, e.g. background jobs, fetching data automatically etc. By doing so, a use-case (Bolt) can not update the trust level of a means without braking functionality. This means MUST only be used to claim the identity of actual and identifying natural persons working under the responsibility of the care organisation, and MUST NOT identify e.g. shared accounts, teams, administrators, dummy or test users.
+
+#### 7.3.1 Attributes
+
+In order for the Nuts node to sign the contract 4 things are needed:
```
woutslakhorst, line 414: In RFCs, we generally refer to *an implementation* and not *Nuts node*

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,81 @@ Beside the mandatory VP fields, the following applies:
...
+
+Note: The _means_ is designed in a way that it is interchangeable with (other) means with a higher trust level, (using a personal authentication device). Even it is technically possible, this means MUST NOT be used to perform machine to machine interactions which are not initiated by the current user, e.g. background jobs, fetching data automatically etc. By doing so, a use-case (Bolt) can not update the trust level of a means without braking functionality. This means MUST only be used to claim the identity of actual and identifying natural persons working under the responsibility of the care organisation, and MUST NOT identify e.g. shared accounts, teams, administrators, dummy or test users.
+
+#### 7.3.1 Attributes
+
+In order for the Nuts node to sign the contract 4 things are needed: 
+
+1. The challenge in the form of a login contract 
+2. The issuing care organisation
```
woutslakhorst, line 417: DID instead of care organization

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,81 @@ Beside the mandatory VP fields, the following applies:
...
+    "@context":[
+      "http://schema.org/",
+      "https://www.w3.org/2018/credentials/v1"
+    ],
+    "type": ["VerifiableCredential"],
+    "issuer":"did:nuts:123456789",
+    "credentialSubject": {
+      "@type": "EmployeeRole",
+      "identifier": "481",
+      "roleName": "Verpleegkundige niveau 2",
```
woutslakhorst, line 437: we can't standardize identifier and roleName yet. This is not part of the current problem and will open up discussions that are not needed at this point.

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,81 @@ Beside the mandatory VP fields, the following applies:
...
+#### 7.3.1 Attributes
+
+In order for the Nuts node to sign the contract 4 things are needed: 
+
+1. The challenge in the form of a login contract 
+2. The issuing care organisation
+3. Personal information about the person (user)
+4. The employee role which contains information about the relationship between the care organisation and the user 
+
+#### 7.3.2 Challenge
```
woutslakhorst, line 421: A credential specification as input is a bit weird. What if we use the presentation exchange format? That would actually also be helpful for any OIDC capable wallet.

rfc/rfc002-authentication-token.md
stevenvegt: I made it an example. Since it is pretty relevant and there are more national things in this spec such as IRMA anyway.

rfc/rfc002-authentication-token.md
stevenvegt: It is the medewerkernummer, which ensures a unique identifier. I improved the spec to have either an unique identifier or an email address. The combination initials and family name is not guaranteed unique. I suspect the employee identifier is probably more available than an email address.
I made the `roleName` optional.

rfc/rfc002-authentication-token.md
stevenvegt: I don't think we want to write a parser for that at the moment. Also, I think this is technically the credential, issued by the care provider which we let the user present by "signing the contract". By adding a `@context`, all these fields have a type and meaning. I used the relational model from schema.org, no custom fields were added.

stevenvegt:
> RFC002 is part of spec v1.0. I think you need to create a new RFC that amends rfc002.

Not sure how a new RFC will solve that problem? I don't like that option since it make thinks harder to find. Also, we explicitly state in section 7.1: 

> Future means will be added when available. 

Which is what we do now.

woutslakhorst:
> > RFC002 is part of spec v1.0. I think you need to create a new RFC that amends rfc002.
> 
> Not sure how a new RFC will solve that problem? I don't like that option since it make thinks harder to find. Also, we explicitly state in section 7.1:
> 
> > Future means will be added when available.
> 
> Which is what we do now.

Well, then we would have to have a conversation about how we create specifications somewhere in the future. RFCs are not meant to change. So we either use RFCs or have versioned specs. But not a mix.

woutslakhorst:
A simplification for the credential could be:

* The `credentialSubject` which contains the following information:
      1. The `@type` which contains the `EmployeeRole` type to indicate the contents describes a EmployeeRole as defined by schema.org
      2. The `identifier` which contains the employee number, username or email. This identifier MUST uniquely identify one natural person within the care organisation.
      3. The `roleName` (optional) which contains the name of the role of the person during this user session.
      4. The `member` which contains the following information:
         1. The `@type` which contains the Person type.
         2. The `initials` which contains the initials of the person.
         3. The `familyName` which contains the last name of the person.

This would remove some if-this-then-that logic. email is really only needed as unique identifier.

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,129 @@ Beside the mandatory VP fields, the following applies:
...
+      1. `http://schema.org/`: contains the Organisation, employeeRole and person types.
+      2. `https://nuts.nl/credentials/v1`: contain definition of the `NutsEmployeeIdentityCredential` type.
+      3. `https://www.w3.org/2018/credentials/v1`: contains definition of verifiable credential fields.
+   * The `type` which MUST contain the `VerifiableCredential` and `NutsEmployeeIdentityCredential` types.
+   * The `issuer` which contains the DID of the care organisation. This DID MUST resolve to a DID Document which contains the public key used to sign the credential.
+   * The `credentialSubject` which contains the following information:
+      1. The `@type` which contains the `EmployeeRole` type to indicate the contents describes a EmployeeRole as defined by schema.org
+      2. The `identifier` (optional) which contains the employee number or username. This identifier MUST uniquely identify one natural person within the care organisation. If this number is not available, the email address field of the `Person` MUST be used.
+      3. The `roleName` (optional) which contains the name of the role of the person during this user session.
```
reinkrul, line 434: Why add this? Is the verifying party supposed to do something with it? If so, is it standardized? (if not, it's probably useless for the verifying party).

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,129 @@ Beside the mandatory VP fields, the following applies:
...
+      4. The `member` which contains the following information:
+         1. The `@type` which contains the Person type.
+         2. The `initials` which contains the initials of the person.
+         3. The `familyName` which contains the last name of the person.
+         4. The `email` (optional) which contains the email address of the person. If the email is not available, the `identifier` field of `EmployeeRole` MUST be present.
```
reinkrul, line 439: Why?

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,129 @@ Beside the mandatory VP fields, the following applies:
 * The `proof.type` field MUST equal `NutsUziSignedContract`.
 * The `proof.proofValue` field MUST contain the JWT in its compact serialization form as described in \[[RFC7515 section 3.1](https://tools.ietf.org/html/rfc7515#section-3.1)\].
 
+### 7.3 NutsEmployeeIdentity
+
+The NutsEmployeeIdentity method is an authentication _means_ which allows a care organisation to assert the identity of the current user.
```
reinkrul, line 406: "a care organization", is this the care organization making a statement on the user's identity, or the party verifying it?

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,129 @@ Beside the mandatory VP fields, the following applies:
...
+The care organisation uses the information of the current user session to present the user with a statement about its identity and asks the user to confirm this.
+After confirmation this bundle of information is singed with the care organisation's private key.
+
+This method makes it possible to assert a user identity without the need for a personal authentication means.
```
reinkrul, line 410: with lower level of assurance (LoA). Now it's only a note, but it's a very important trait of this authentication means.

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,129 @@ Beside the mandatory VP fields, the following applies:
...
+#### 7.3.3 Response
+
+The response MUST contain a signed VerifiableCredential with a signature from the issuing care organisation.
+This credential is contained in a VerifiablePresentation. 
+The presentation contains a proof with a signature.
+Since the user doesn't have a private key, the private key from the care organisation is also used to sign the presentation.
+The proof section MUST contain the `challenge` from the request.
+
+To add further accountability, an `evidence` SHOULD be added.
+This object MUST be of the type `NutsBrowserSession` which SHOULD contain the `user-agent` field and the `ipAddress` field.
```
reinkrul, line 481: why?

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,129 @@ Beside the mandatory VP fields, the following applies:
...
+  }
+}
+```
+
+#### 7.3.3 Response
+
+The response MUST contain a signed VerifiableCredential with a signature from the issuing care organisation.
+This credential is contained in a VerifiablePresentation. 
+The presentation contains a proof with a signature.
+Since the user doesn't have a private key, the private key from the care organisation is also used to sign the presentation.
```
reinkrul, line 477: which one?

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,129 @@ Beside the mandatory VP fields, the following applies:
...
+        "initials": "J",
+        "familyName": "van Dijk",
+        "email": "j.vandijk@example.com"
+      }
+    }
+  }
+}
+```
+
+#### 7.3.3 Response
```
reinkrul, line 472: What should a verifier check, on the response?

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,129 @@ Beside the mandatory VP fields, the following applies:
 * The `proof.type` field MUST equal `NutsUziSignedContract`.
 * The `proof.proofValue` field MUST contain the JWT in its compact serialization form as described in \[[RFC7515 section 3.1](https://tools.ietf.org/html/rfc7515#section-3.1)\].
 
+### 7.3 NutsEmployeeIdentity
+
+The NutsEmployeeIdentity method is an authentication _means_ which allows a care organisation to assert the identity of the current user.
+The care organisation uses the information of the current user session to present the user with a statement about its identity and asks the user to confirm this.
```
reinkrul, line 407:
```suggestion
The issuing care organisation uses the information of the current user session to present the user with a statement about its identity and asks the user to confirm this.
```

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,129 @@ Beside the mandatory VP fields, the following applies:
 * The `proof.type` field MUST equal `NutsUziSignedContract`.
 * The `proof.proofValue` field MUST contain the JWT in its compact serialization form as described in \[[RFC7515 section 3.1](https://tools.ietf.org/html/rfc7515#section-3.1)\].
 
+### 7.3 NutsEmployeeIdentity
```
reinkrul, line 404: UZI and IRMA means are called just so, `NutsEmployeeIdentity` is the name of the credential. I think the means should be called something along the line of "Self-claimed Employee Identity" or just "Employee Identity".


Also; should be now add levels of assurances to the methods we spec'd? (or refer to a framework we use, e.g. eIDAS(2)?)

reinkrul (changes requested):

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,129 @@ Beside the mandatory VP fields, the following applies:
...
+
+#### 7.3.1 Attributes
+
+In order for the implementation to sign the contract the following information is needed:
+
+1. The `challenge` in the form of a login contract.
+2. The `credential` which contains the following information:
+   * The `@context` which MUST contain the following urls:
+      1. `http://schema.org/`: contains the Organisation, employeeRole and person types.
+      2. `https://nuts.nl/credentials/v1`: contain definition of the `NutsEmployeeIdentityCredential` type.
```
reinkrul, line 427: actually a bit of an oddball VC since it's a VC for an employee, but comes from the wallet of the care organization (rather than a personal wallet). It's sort of a shortcut since the employee doesn't have an SSI wallet with an EmployeeIdentity credential?

rfc/rfc002-authentication-token.md
```diff
@@ -401,3 +401,129 @@ Beside the mandatory VP fields, the following applies:
...
+      3. `https://www.w3.org/2018/credentials/v1`: contains definition of verifiable credential fields.
+   * The `type` which MUST contain the `VerifiableCredential` and `NutsEmployeeIdentityCredential` types.
+   * The `issuer` which contains the DID of the care organisation. This DID MUST resolve to a DID Document which contains the public key used to sign the credential.
+   * The `credentialSubject` which contains the following information:
+      1. The `@type` which contains the `EmployeeRole` type to indicate the contents describes a EmployeeRole as defined by schema.org
+      2. The `identifier` (optional) which contains the employee number or username. This identifier MUST uniquely identify one natural person within the care organisation. If this number is not available, the email address field of the `Person` MUST be used.
```
reinkrul, line 433: Maybe this should be in the "evidence" field?

rfc/rfc002-authentication-token.md
woutslakhorst: Unless you give the employee a ephemeral wallet (did:key)?

PR #257, opened 2023-04-12
RFC018: Issuing credentials using OIDC4VCI

reinkrul:
TODO:

- [x] Abstract
- [x] Specify custom CredentialOffer error response (introduced by https://github.com/nuts-foundation/nuts-node/pull/2207) 

Fixes https://github.com/nuts-foundation/nuts-specification/issues/258


rfc/rfc018-oidc4vci-credential-issuance.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Copyright Notice
+
+![](../.gitbook/assets/license.png)
+
+This document is released under the [Attribution-ShareAlike 4.0 International \(CC BY-SA 4.0\) license](https://creativecommons.org/licenses/by-sa/4.0/).
+
+## 1. Introduction
+
+This specification describes how to apply OpenID for Verifiable Credential Issuance (OIDC4VCI) to issue Verifiable Credentials (VCs)
```
woutslakhorst, line 24:
```suggestion
This specification describes how to apply OpenID Connect for Verifiable Credential Issuance (OIDC4VCI) to issue Verifiable Credentials (VCs)
```

rfc/rfc018-oidc4vci-credential-issuance.md
```diff
@@ -0,0 +1,84 @@
...
+
+## 2. Terminology
+
+This specification uses the terms "DID", "DID document" and "Service" as defined by the [DID Core Data Model](https://www.w3.org/TR/did-core/),
+"Verifiable Credential", "Credential Subject" and "Issuer" as defined by the [Verifiable Credential Data Model](https://www.w3.org/TR/vc-data-model/), and
+"Wallet", "Client Metadata" and "Credential Offer" as defined by [OpenID 4 Verifiable Credential Issuance](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html).
+
+## 3. Overview
+
+While OIDC4VCI is typically designed to be used to issue credentials through a browser or to a mobile wallet app,
```
woutslakhorst, line 35: assumption? It mainly **describes** issuance to a mobile wallet.

woutslakhorst (changes requested):
I think it also misses the fact that an implementation MUST implement both the wallet and issuer roles.

rfc/rfc018-oidc4vci-credential-issuance.md
gerardsn: I'm not sure where the `oidc4vci` comes from, but I propose to replace it with `OpenID4VCI` (so without `Connect` since that is not in the original spec, and same for `OpenID4VP`)
This is a lot easier to pronounce, and seems to be the preferred reference in the specs and it is also how they are referred to from the ARF

rfc/rfc018-oidc4vci-credential-issuance.md
```diff
@@ -0,0 +1,84 @@
...
+implementations MUST NOT require a PIN in server-to-server flows (see Appendix A.1).
+
+## 4. Client Metadata Discovery using DID documents
+
+To offer a credential the issuer needs to discover the wallet's credential offer endpoint, which is registered in the client metadata.
+
+When the wallet is represented by a resolvable DID the client should register a `oidc4vci-wallet-metadata` service on the DID document,
+which contains a URL pointing to the client metadata. This indicates the wallet supports receiving credentials according to the OIDC4VCI specification.
+
+The issuer can resolve this URL and retrieve the client metadata to find the credential offer endpoint.
```
gerardsn, line 51: Missing something about this being the only way (or not) a credential is offered when the service exists

rfc/rfc018-oidc4vci-credential-issuance.md
```diff
@@ -0,0 +1,84 @@
...
+## Appendix A: Design decisions
+
+### A.1 No PIN requirement
+
+OIDC4VCI specifies a PIN to mitigate credential stealing,
+e.g. someone looking over the shoulder of the intended holder standing at a terminal,
+quickly scanning the QR code and stealing the resulting credential.
+Since this specification describes server-to-server flows, there is no user involved from who the QR code can be stolen
+or coerced into giving the QR code.
+To prevent eavesdropping, HTTPS MUST be used to protect communication (see Appendix B.1).
```
gerardsn, line 68: imo appendices should not introduce requirements. The document should be self-contained and the appendices can include examples or information to clarify the document. So appendix B should probably be section 6

gerardsn (changes requested):

reinkrul:
> I think it also misses the fact that an implementation MUST implement both the wallet and issuer roles.

why?

woutslakhorst:
> > I think it also misses the fact that an implementation MUST implement both the wallet and issuer roles.
> 
> why?

Otherwise it's not a Nuts node, just a half one.

rfc/rfc018-oidc4vci-credential-issuance.md
```diff
@@ -0,0 +1,84 @@
...
+| Nuts foundation           | R.G. Krul |
+| Request for Comments: 018 | Nedap     |
+
+## Verifiable Credential Issuance using OpenID
+
+### Abstract
+
+### Status of document
+
+This document describes a Nuts standards protocol.
```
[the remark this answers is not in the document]
reinkrul: not sure

rfc/rfc018-openid-credential-issuance.md
```diff
@@ -0,0 +1,144 @@
...
+
+Nodes MUST support both processes of issuing and receiving credentials over OpenID4VCI.
+
+## 3.1 Supported Flows
+
+Issuers MUST support the pre-authorized code flow to issue credentials.
+While OpenID4VCI specifies a PIN to mitigate credential stealing,
+implementations MUST NOT require a PIN in server-to-server flows (see Appendix A.1).
+
+The following from the OpenID4VCI specification does not need to be supported:
```
woutslakhorst, line 50: "does not need" might be too vague.
Are all the items in the list restricted by what's advertised in the (wallet/issue) metadata? If not then the does not becomes a must not?

rfc/rfc018-openid-credential-issuance.md
```diff
@@ -0,0 +1,144 @@
...
+
+## 3.2 Credential Offer Response
+
+The OpenID4VCI specification does not specify responses for the credential offer.
+But since this specification describes server-to-server interaction
+a response is useful for the issuer to know whether the credential offer was accepted.
+
+### 3.2.1 Credential Offer Success Response
+
+In case the wallet accepted the offer and will try to retrieve the credential, it MUST respond with HTTP status `200 OK`.
```
woutslakhorst, line 69: You mean retrieve within the context of the offer call? Not async

rfc/rfc018-openid-credential-issuance.md
```diff
@@ -0,0 +1,144 @@
...
+### A.1 No PIN requirement
+
+OpenID4VCI specifies a PIN to mitigate credential stealing,
+e.g. someone looking over the shoulder of the intended holder standing at a terminal,
+quickly scanning the QR code and stealing the resulting credential.
+Since this specification describes server-to-server flows, there is no user involved from who the QR code can be stolen
+or coerced into giving the QR code.
+To prevent eavesdropping, HTTPS MUST be used to protect communication (see Appendix B.1).
+
+### A.2 Trust issuer based on TLS certificates
```
woutslakhorst, line 138: Do we do this currently?

rfc/rfc018-openid-credential-issuance.md
reinkrul: Then you can't have a generic OpenID4VCI wallet that also support Nuts, since a compelte implementation vs this RFC are mutually exclusive.

Maybe we should inverse it, specifying what MUST be supported? Although I wrote it up this way since that sounded easier.

rfc/rfc018-openid-credential-issuance.md
reinkrul: No, just whether the wallet accepted the credential offer. I'll clarify the sentence.

rfc/rfc018-openid-credential-issuance.md
reinkrul: I think I tried to convey that we only trust issuers from trusted certificates, which is defined by the truststore. And since identifier URLs are HTTPS URLs, they're trusted if the connection can be made.

rfc/rfc018-openid-credential-issuance.md
woutslakhorst: Usually I scan to see what I MUST do and skip everything else ;)

woutslakhorst (commented):
I think the does not needs to be inversed to MUSTs

woutslakhorst (approved):

rfc/rfc018-openid-credential-issuance.md
```diff
@@ -0,0 +1,140 @@
...
+Nodes MUST support both processes of issuing and receiving credentials over OpenID4VCI.
+
+## 3.1 Supported Flows and Features
+
+Conforming implementations MUST support the following features:
+
+- Pre-authorized code flow
+- Credential offer:
+  - with `credential_offer` parameter
+  - with `string` values, referring to a credential's `id` in `credentials_supported`
```
gerardsn, line 51: We currently do not support this since our credential definitions do not contain an `id`? @woutslakhorst

rfc/rfc018-openid-credential-issuance.md
```diff
@@ -0,0 +1,140 @@
...
+- Credentials in JSON-LD format
```
gerardsn, line 52: == `ldp_vc` format type

rfc/rfc018-openid-credential-issuance.md
```diff
@@ -0,0 +1,140 @@
...
+## 3.2 Credential Offer Response
+
+The OpenID4VCI specification does not specify responses for the credential offer.
+But since this specification describes server-to-server interaction
+a response is useful for the issuer to know whether the credential offer was accepted.
+
+### 3.2.1 Credential Offer Success Response
+
+In case the wallet considers the offer valid, it MUST respond with HTTP status `200 OK`.
+This is regardless of whether the wallet will actually revrieve the wallet.
```
gerardsn, line 64:
```suggestion
This is regardless of whether the wallet will actually retrieve the credential.
```

rfc/rfc018-openid-credential-issuance.md
```diff
@@ -0,0 +1,140 @@
...
+
+If the wallet decides to retrieve the credential synchronously, it SHOULD respond with `status` set to `credential_received`.
```
gerardsn, line 66: OpenID4VCI spec has renamed the **credential request** `synchronous` flow to `immediate issuance` flow. We probably want to use the same terminology? But I do not know how/if this applies to the offer...

rfc/rfc018-openid-credential-issuance.md
```diff
@@ -0,0 +1,140 @@
...
+  "error": "invalid_grant"
+}
+```
+
+## 4. Client Metadata Discovery using DID documents
+
+To offer a credential the issuer needs to discover the wallet's credential offer endpoint, which is registered in the client metadata.
+
+To allow a (Nuts) wallet to be discovered, the wallet MUST be represented by a resolvable DID.
+The DID document MUST contain a `node-http-services-baseurl` service, which contains the base URL for node-to-node HTTP services.
```
gerardsn, line 105:
```suggestion
The DID document MUST contain a `node-http-services-baseurl` service type, which contains the base URL for node-to-node HTTP services.
```

rfc/rfc018-openid-credential-issuance.md
```diff
@@ -0,0 +1,140 @@
...
+
+### 5.1 Credential Issuer Trust
+
+In server-to-server issuance of credentials, there's no user approving the credential offer before requesting it and loading it into the holder's wallet.
+Therefore, the holder must make a decision whether to trust the credential issuer.
+Trust SHOULD be derived from the TLS trust anchors referred to by section 5.2 (Transport Security).
+
+### 5.2 Transport Security
+
+To prevent against eavesdropping and man-in-the-middle attacks, all exchanges MUST be done over HTTPS.
```
gerardsn, line 119: This is already a hard requirement derived from the OAuth/OpenID specs that the OpenID4VCI spec builds on. Maybe we should mention this since this spec only defines what we added?

rfc/rfc018-openid-credential-issuance.md
```diff
@@ -0,0 +1,140 @@
...
+
+Nodes MUST support both processes of issuing and receiving credentials over OpenID4VCI.
+
+## 3.1 Supported Flows and Features
+
+Conforming implementations MUST support the following features:
+
+- Pre-authorized code flow
+- Credential offer:
+  - with `credential_offer` parameter
```
gerardsn, line 50: we use `credential_offer_uri`, while our OAPI specs describe the `credential_offer` scheme

rfc/rfc018-openid-credential-issuance.md
```diff
@@ -0,0 +1,140 @@
...
+or coerced into giving the QR code.
+To prevent eavesdropping, HTTPS MUST be used to protect communication (see Appendix B.1).
+
+### A.2 Trust issuer based on TLS certificates
+
+Since the TLS trust anchors are curated, trusting credential issuers can be derived from trust TLS server certificates:
+Credential issuer identifiers are URLs and thus link to a trusted TLS certificate.
+
+Future, more involved trust models could involve the issuer presenting a credential from which trust is derived.
+For instance, proving that the issuer is a registered legal entity, which can be held accountable.
```
gerardsn, line 140: none of us are lawyers, so I would refrain from statements like this

rfc/rfc018-openid-credential-issuance.md
gerardsn: For the `credentials` array I think we only support the `JSON object` containing the `ldp_vc` format specific `credential_definition` as defined for in https://openid.bitbucket.io/connect/editors-draft/openid-4-verifiable-credential-issuance-1_0.html#name-vc-secured-using-data-integ

rfc/rfc018-openid-credential-issuance.md
```diff
@@ -0,0 +1,140 @@
...
+
+## 3.2 Credential Offer Response
+
+The OpenID4VCI specification does not specify responses for the credential offer.
+But since this specification describes server-to-server interaction
+a response is useful for the issuer to know whether the credential offer was accepted.
+
+### 3.2.1 Credential Offer Success Response
+
+In case the wallet considers the offer valid, it MUST respond with HTTP status `200 OK`.
```
gerardsn, line 63: Just to check, we still need to implement this right?

gerardsn: This conflicts with the synchronous flow. An invalid credential in the synchronous flow for a valid offer would result in a 400 while this line says that it MUST be a 200. A better separation between sync/async flow should fix this

gerardsn (changes requested):

rfc/rfc018-openid-credential-issuance.md
reinkrul: It's a slightly different thing here, from the wallet's perspective. This is not spec'd. But I renamed it to be more like the spec.

rfc/rfc018-openid-credential-issuance.md
reinkrul: the DID document contains a service with that type, not the service type

rfc/rfc018-openid-credential-issuance.md
reinkrul: We use the credential_offer parameter, which in encoded in the credential offer request (issuer->wallet). Using URI would mean the issuer provides a URI at which the wallet can retrieve the credential.

rfc/rfc018-openid-credential-issuance.md
reinkrul: It says when it is considered valid? So if valid, 200, if not 400.

The sync part will probably change when we support retying.

gerardsn: I guess it depends on what you consider to be "the offer". My interpretation was that this is just the `credential_offer` message, but seems that you refer to the entire flow which includes retrieval of the credential for the sync flow.

rfc/rfc018-openid-credential-issuance.md
gerardsn: don't know what I was thinking here

rfc/rfc018-openid-credential-issuance.md
gerardsn: Recent versions of the spec seem to have replaced `id` by `scope`. The `scope` field is also not present in our definitions. Not sure if the node or this RFC should be changed?

reinkrul:
We're not going to support this as it currently is

PR #267, opened 2023-10-22
Discovery Service RFC

reinkrul:
Things to decide on:
- Do we want to use PEX to determine what credentials to present? Otherwise, implementations will have to hardcode how to build presentations (with which credentials) for each list. It can then also be used to validate presentations to be added to the list. Presentation definitions can be distributed out-of-band, or hosted by the list (e.g. query `/<list-endpoint>/metadata.json`) -> decided have a use case definitions, created by use case maintainers and shared with all users of the use case
- The list is now just that: a list of presentations. It's hard for clients to see which entries changed (were added/removed), which might be useful since checking the presentations might become (more) expensive when the list grows (e.g. entries might be converted to a SQL table for searchability). But recognizing presentations is hard, options: -> decided to let clients query deltas
  - Use presentation.ID, but guaranteeing uniqueness is hard/impossible (since it's set by the presentation creator)
  - Use hash of the presentation (JWT can be hashed as-is, for JSON-LD maybe use proof hash/signature?)


woutslakhorst:
- yes we want to use PEX for credential selection. It is distributed by the use case.
- use `jwt_vc` and check the `jti`, `iat` and `exp`. If still valid check DB for `jti`, if not present, check signature.

rfc/rfc0XX-use-case-list.md
```diff
@@ -0,0 +1,139 @@
...
+The list endpoint is an HTTP endpoint exposed by the list maintainer.
+It acts like a REST resource, which can be read like a list and updated to by sending a Verifiable Presentation to it.
+
+### 3.1 Format
+
+The list is a JSON array containing zero or more Verifiable Presentations.
+List entries can be a:
+
+- presentation in JWT format as JSON string.
+- presentation in JSON-LD format as JSON object.
```
woutslakhorst, line 43: just one format will be enough for now.

rfc/rfc0XX-use-case-list.md
```diff
@@ -0,0 +1,139 @@
...
+
+### 3.2 Registering on the list
+
+Clients can register a Verifiable Presentation on the list by sending an HTTP POST to the maintainer's list endpoint.
+The HTTP request body MUST be a Verifiable Presentation encoded as JSON. The content type of the request MUST be `application/json`.
```
woutslakhorst, line 48: I would say MUST be a jwt

rfc/rfc0XX-use-case-list.md
```diff
@@ -0,0 +1,139 @@
...
+
+The maintainer MUST validate the Verifiable Presentation as specified in section 4. If the validation fails, it MUST return a 400 Bad Request response.
+If the validation succeeds, the Verifiable Presentation MUST be added to the list and the maintainer MUST return a 201 Created response.
+Credential subjects on the list MUST be unique, so if the subject already exists, the new presentation replaces the old one.
```
woutslakhorst, line 52: definition of unique? presenter and list of VCs?

woutslakhorst:
Remove JSON-LD support since the semantic meaning is not used. There's only a check against the PD. This could also allow for forward compatibility. JSON-LD will definitely not be forward compatible.

woutslakhorst (commented):
Wouldn't it be better to replace `state` with timestamps or a lamport clock?
Timestamps don't require any additional storage but suffer from nanosecond race conditions. A lamport clock doesn't suffer from this, any new POST is just a +1. This could also help with a tombstone list: [~57, 59, 60], up to 57 are removed as well as 59 and 60.

rfc/rfc0XX-use-case-list.md
```diff
@@ -90,20 +90,24 @@ If one or more presentations are not valid, it SHOULD NOT reject the other prese
...
-Clients MUST re-submit their registration before that period of time has passed, if they want to keep their presentation on the list (once a day, for instance).
+Clients that wish to retain their presentation MUST re-submit their registration before that period of time has passed, if they want to keep their presentation on the list (once a day, for instance).
 
-Maintainers MUST remove presentations which ``expirationDate`` has passed from the list.
+Maintainers SHOULD remove presentations which contains credentials that have been revoked.
+
+### 3.5. Removing entries
+
+TODO: Clients may remove a credential subject by performing an HTTP DELETE to the list endpoint, where the body contains an empty Verifiable Presentation signed by the credential subject.
+TODO: If we support this: we need to serve a tombstone set as well (and also for revocations)
```
woutslakhorst, line 102: If you add a lamport clock per entry and a tombstone list then you have the feature complete.

rfc/rfc0XX-use-case-list.md
```diff
@@ -90,20 +90,24 @@ If one or more presentations are not valid, it SHOULD NOT reject the other prese
...
 
 ## 4. Presentation processing
 
 To process a presentation, the following validation steps MUST be performed:
 
-- ``issuanceDate`` of the presentation MUST have passed.
-- ``expirationDate`` of the presentation MUST NOT have passed.
-- ``expirationDate`` MUST be after ``issuanceDate``.
+- ``nbf`` (not before) of the presentation MUST have passed.
```
woutslakhorst, line 108: I think iat is usually preferred over nbf.

rfc/rfc0XX-use-case-list.md
```diff
@@ -128,6 +128,7 @@ Presentations and credentials MUST be in JWT format.
 
 Maintainers MUST share a JSON document describing the list. This document is known as the _list definition_.
 The document MUST contain the following properties:
+- `id` REQUIRED. JSON string containing a value identifies the list. MAY be used to version the definition.
```
woutslakhorst, line 148: call the `JSON strings` just `string`

woutslakhorst (approved):

woutslakhorst (changes requested):
missing security and privacy considerations.

gerardsn (changes requested):
Normally the response to a Presentation Definition is a (list of) VP(s) + Presentation Submission. We're missing the PS here.

I got a bit carried away during reviewing, sorry

rfc/rfc0XX-use-case-list.md
```diff
@@ -0,0 +1,221 @@
...
+### 3.3 Reading the list
+
+Clients MUST load the list by sending an HTTP GET to the maintainer's list endpoint.
+The maintainer MUST return a 200 OK response with a JSON object containing:
+- `entries` REQUIRED. MUST contain a JSON array with zero or more presentations.
+- `tombstones` REQUIRED. MUST contain a JSON array with zero or more presentation IDs that have been removed from the list.
+- `timestamp` REQUIRED. MUST be a JSON number containing the Lamport timestamp of the last entry.
+
+The `timestamp` query parameter MAY be used by the client to request a delta next time it reads the list.
+If no `timestamp` query parameter is provided, the maintainer MUST return the full list.
```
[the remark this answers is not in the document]
reinkrul: first available LC, which is also the entire list. Due to the rather short-livedness of presentations the list will stay rather short by nature

rfc/rfc0XX-use-case-list.md
```diff
@@ -0,0 +1,221 @@
...
+
+The Verifiable Presentation MUST NOT be valid longer than the Verifiable Credentials it contains. 
+
+### 3.3 Reading the list
+
+Clients MUST load the list by sending an HTTP GET to the maintainer's list endpoint.
+The maintainer MUST return a 200 OK response with a JSON object containing:
+- `entries` REQUIRED. MUST contain a JSON array with zero or more presentations.
+- `tombstones` REQUIRED. MUST contain a JSON array with zero or more presentation IDs that have been removed from the list.
+- `timestamp` REQUIRED. MUST be a JSON number containing the Lamport timestamp of the last entry.
```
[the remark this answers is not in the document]
reinkrul: Could be a implementation suggestion in this RFC

jorritspee:
Sorry to enter the dialogue late and without proper preparation. My question is whether you guys see a relation between the use case list and the DCAT-standard or not? DCAT is mentioned often in relation to FAIR-datapoints.

reinkrul:
> Sorry to enter the dialogue late and without proper preparation. My question is whether you guys see a relation between the use case list and the DCAT-standard or not? DCAT is mentioned often in relation to FAIR-datapoints.

I had to look up DCAT, but I understand it describes the models of the data sets. The goal of this RFC is to define a way to discover parties (and ultimately services) supporting a certain feature/use case, but not how information/data exchanged within that use case is structured/modeled.

I could imagine use case definitions could include URLs/references to a schema or catalog describing the information model of the use case. On the other hand, the Nuts node does not deal with the actual information itself (which is exchanged directly between the XIS's). But, if the use case definition would be used in other systems (than the Nuts node), it might have some value?

rfc/rfc0XX-use-case-list.md
```diff
@@ -0,0 +1,221 @@
...
+Clients that wish to retain their presentation MUST re-submit their registration before that period of time has passed, if they want to keep their presentation on the list (once a day, for instance).
+
+Maintainers SHOULD remove presentations which contains credentials that have been revoked.
+
+### 3.5. Removing presentations
+
+Clients can remove presentations from a specific credential subject from a list, e.g. when a care organization stops supporting the use case.
+To remove presentations of a credential subject the client MUST send an HTTP DELETE request to the list endpoint,
+where the body contains an empty Verifiable Presentation signed by the credential subject.
+Any credentials in the presentation MUST be ignored by the maintainer.
```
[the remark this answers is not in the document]
reinkrul: Will replace it with a VP that specifies the ID of the VP being removed.

jorritspee:
> > Sorry to enter the dialogue late and without proper preparation. My question is whether you guys see a relation between the use case list and the DCAT-standard or not? DCAT is mentioned often in relation to FAIR-datapoints.
> 
> 
> 
> I had to look up DCAT, but I understand it describes the models of the data sets. The goal of this RFC is to define a way to discover parties (and ultimately services) supporting a certain feature/use case, but not how information/data exchanged within that use case is structured/modeled.
> 
> 
> 
> I could imagine use case definitions could include URLs/references to a schema or catalog describing the information model of the use case. On the other hand, the Nuts node does not deal with the actual information itself (which is exchanged directly between the XIS's). But, if the use case definition would be used in other systems (than the Nuts node), it might have some value?

Hi Rein,
Thanks for your reply. DCAT is generally used by a partner to publish what data sets and/or - services a party provides. I would think that a use case could map to a data service. @hietkamp, you also had some ideas on using DCAT in a Nuts context didn't you?

reinkrul:
Renamed to Discovery Service with Client and Server, removed tombstone set in favor of "just" a Verifiable Presentation that refers a retracted Verifiable Presentation.

hietkamp:
We plan to agree on a data catalog (dcat2) with KIK-V. This would be a service address in Nuts to an endpoint where the data catalog can be found. The data catalog is intended to provide information about the datasets you offer and in accordance with which standard you offer this data (the ontology in KIK-V).

I myself see more value in mentioning the ontology (you can call it a domain model) that you support. You make data available in accordance with an (version of an) ontology. The principle is that we want to separate data and functionality.

An ontology is suitable for a use case, that's what you agree on. But the ontology can also be used for many other use cases. The use case is also owned by the data user and not the data holder in my experience. Then it is strange that the holder specifies the use cases for which his data is suitable.

The RFC therefore plays into the use case-driven way of working while we want to move towards a data-driven way of working. We may still need this for the first plateau of the NVS, but then we will immediately build an architectural debt towards the second plateau of the NVS.

woutslakhorst:
> Then it is strange that the holder specifies the use cases for which his data is suitable.

A lot of the current development for data exchange is use-case oriented and even defines a specific information model per use-case. If all those models converge in the future, there indeed could be a 'single use-case' that only requires a party to publish its data catalog endpoint. This discovery service would then still be useable. It would be a single use-case, single instance usage. A way to discover other endpoints will always be required.

rfc/rfc0XX-discovery-service.md
```diff
@@ -0,0 +1,232 @@
...
+
+To only keep valid presentations on the list, servers MUST remove presentations whose `exp` (expiration) time has passed from the list.
+
+Clients that wish to retain their presence on the list MUST submit a new registration before the current entry's `exp` time has passed (once a day, for instance).
+
+Servers SHOULD remove presentations which contain credentials that have been revoked.
+
+### 3.4 Retracting presentations
+
+Clients can remove presentations from the list, e.g. when a care organization stops supporting the use case.
```
woutslakhorst, line 104: RFCs are independent of the domain

woutslakhorst (changes requested):

gerardsn (changes requested):
Introduction, security, and privacy considerations are missing.

rfc/rfc0XX-discovery-service.md
```diff
@@ -0,0 +1,232 @@
...
+To remove presentations of a credential subject the client MUST send an HTTP DELETE request to the service endpoint,
```
reinkrul, line 105: Note to self: just use the same HTTP POST, only with different body

woutslakhorst:
@reinkrul is the name/id of the use case included in the VP? Otherwise you can replay a VP to another use case....

reinkrul:
Renamed "timestamp" to "tag", as it is an opaque value. "Timestamp" suggests it's a time-based property, which it might not be.

reinkrul:
> @reinkrul is the name/id of the use case included in the VP? Otherwise you can replay a VP to another use case....

This is now included;

> - `aud` MUST contain the ID of the service.

reinkrul:
> [JWT-VC profile](https://identity.foundation/jwt-vc-presentation-profile/#nbf-jwt-claim)
>
>[VC Data Model v1.1](https://identity.foundation/jwt-vc-presentation-profile/#term:vc-data-model-v1.1) specifies that “issuanceDate” property MUST be represented as an nbf JWT claim, and not iat JWT claim. This might sound couterintuitive, but the implementers of this profile MUST follow this guidance.
>
>Do we need a similar statement? (If only to prevent reconsidering/discussions in the future)

If we already refer to VC 1.1, it's covered? (Since it's a MUST)

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,252 @@
...
+### 3.1 Registration
+
+Clients can publish a Verifiable Presentation to the list by sending it in an HTTP POST to the service endpoint.
+The HTTP request body MUST be a Verifiable Presentation (in JWT format). The content type of the request MUST be `application/json`.
+
+The server MUST validate the Verifiable Presentation as specified in section 4. If the validation fails, it MUST return a 400 Bad Request response.
+If the validation succeeds, the Verifiable Presentation MUST be added to the list and the server MUST return a 201 Created response.
+
+A credential subject (identified by `credentialSubject.id`) MUST NOT appear more than once on the list,
+so a new registration MUST replace the previous one from the same credential subject.
```
woutslakhorst, line 60: `the list` probably refers to §3.2. This is confusing for the uninitiated

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,252 @@
...
+``
+
+The Verifiable Presentation MUST NOT be valid longer than the Verifiable Credentials it contains. 
+
+### 3.2 Reading
+
+Clients MUST read the list by sending an HTTP GET to the server's service endpoint.
+The server MUST return a 200 OK response with a JSON object containing:
+- `entries` REQUIRED. MUST contain a JSON array with zero or more presentations.
+- `tag` REQUIRED. MUST be a JSON string referring to the last entry. The tag is an opaque value; the client SHOULD NOT derive any meaning from it.
```
woutslakhorst, line 78: > referring to the last entry.

It refers to the moment in time of the last entry (in the ref impl). A description like: "used by the server to provide delta's" might be better suited

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,252 @@
...
+  "entries": [
+    "ey1234.etc.etc",
+    "ey5678.etc.etc"
+  ],
+  "tag": "515",
+}
+``
+
+Clients MUST validate each presentation in the list as specified in section 4.
+If a presentation is valid, the client uses it in its system. If a presentation is not valid, it MUST be rejected.
```
woutslakhorst, line 100:
```suggestion
If a presentation is not valid, it MUST be rejected.
```

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,252 @@
...
+If one or more presentations are not valid, it SHOULD NOT reject the other presentations.
+
+### 3.3 List pruning
```
woutslakhorst, line 103: How does this work with expired (not retracted) VPs? Do clients also have to prune their lists?

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,252 @@
...
+To process a presentation, the following validation steps MUST be performed:
+
+- `jti` (JWT ID) of the presentation MUST be a non-empty string.
+- the presentation ID  (`jti`) MUST NOT already exist on the list.
+- `nbf` (not before) of the presentation MUST have passed.
+- `exp` (expiration) of the presentation MUST NOT have passed.
+- `exp` MUST be after `nbf`.
+- `aud` MUST contain the ID of the service.
+- the number of seconds between `nbf` and `exp` MUST NOT exceed `presentation_max_validity` (see Service Definition).
+- all credential issuers MUST be trusted (see section 5).
```
woutslakhorst, line 141: implicit through Presentation Definition?

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,252 @@
...
+        }
+      }
+    ]
+  }
+}
+``
+
+## 6. Trust
+
+Trust of credential issuers (e.g. `did:example:education-accredetor` issuing `EducationalInstitutionCredential`) SHOULD be defined by the presentation definition.
```
woutslakhorst, line 204: `accreditor`

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,252 @@
...
+    "filter": {
+      "type": "string",
+      "const": "EducationalInstitutionCredential"
+    }
+  },
+  {
+    "path": "$.issuer",
+    "filter": {
+      "type": "string",
+      "const": "did:example:education-accredetor"
```
woutslakhorst, line 220: `accreditor`

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,252 @@
...
+    }
+  }
+]
+``
+
```
woutslakhorst, line 225: Missing security and privacy considerations

SUMMARY.md
```diff
@@ -20,6 +20,7 @@
 * [RFC019 Employee Identity Authentication Means](rfc/rfc019-employee-identity-means.md)
 * [RFC020 Authorization credential extension](rfc/rfc020-authorization-credential-extension.md)
 * [RFC021 VP Token Grant Type](rfc/rfc021-vp_token-grant-type.md)
+* [RFC022 Discovery Service](rfc/rfc022-discovery-service.md)
```
woutslakhorst, line 23: Also update graph image

woutslakhorst (changes requested):

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,252 @@
...
+That way, timestamps unambiguously reference the last presentation the client received.
+
+A Unix timestamp does not offer this property, as it is possible that multiple presentations are received at the second,
+possibly serving clients duplicate presentations.
+
+## References
+
+* [JWT] M. Jones, J. Bradley, N. Sakimura, "JSON Web Token (JWT)", RFC 7519, May 2015, <https://www.rfc-editor.org/info/rfc7519>.
+* [VC-DATA-MODEL] M. Sporny, D. Longley, D. Chadwick, "Verifiable Credentials Data Model 1.1", W3C Recommendation, 3 March 2022, <https://www.w3.org/TR/vc-data-model/>.
+* [PE] D. Buchner, B. Zundel, M. Riedel, K.H. Duffy, "Presentation Exchange 2.0.0", 12 September 2023, <https://identity.foundation/presentation-exchange/spec/v2.0.0/>.
```
gerardsn, line 251: This is usually referred to as `PEX` since that sounds a lot nicer when pronounced.

gerardsn (changes requested):
A credential is not the same as a verifiable credential. The same goes for presentation/VP. I'd prefer to add VP and VC to the terminology section and replace all.

woutslakhorst (changes requested):
missing error return

woutslakhorst (approved):

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,299 @@
...
+  the JWT `kid` header MUST reference an `assertionMethod` key from the subject's DID document.
+- The DID of the credential subject MUST be of a DID method defined in the service definition.
+- the presentation MAY contain a `DiscoveryRegistrationCredential` credential defined by the Nuts JSON-LD context (https://nuts.nl/credentials/v1).
+  If present, it MUST NOT be included in the next validation steps.
+- the credentials MUST conform to the Presentation Definition associated with the list (see Service Definition).
+  The Verifiable Presentation MUST NOT contain any Verifiable Credentials besides the ones that conform to the Presentation Definition and the `DiscoveryRegistrationCredential`.
+
+If a validation step fails, the presentation MUST be rejected.
+
+JWT credential and presentation encoding as specified by [VC-DATA-MODEL] MUST be applied.
```
gerardsn, line 164: is this saying that credentials also MUST be JWT?

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,299 @@
...
+A clock skew of 5 seconds MAY be applied to `nbf` and `exp` claims.
+
+### 4.1 DiscoveryRegistrationCredential
+
+The optional `DiscoveryRegistrationCredential` is used to add additional parameters to the registration.
+It's a _holder_ credential, meaning it has been self issued. Fields under `credentialSubject` (except `id`) can be used by clients depending on the use case.
+The credential has the following requirements:
+- `type` MUST be `DiscoveryRegistrationCredential`.
```
gerardsn, line 172:
```suggestion
- `type` MUST contain `DiscoveryRegistrationCredential`.
```

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,299 @@
...
+  If present, it MUST NOT be included in the next validation steps.
+- the credentials MUST conform to the Presentation Definition associated with the list (see Service Definition).
+  The Verifiable Presentation MUST NOT contain any Verifiable Credentials besides the ones that conform to the Presentation Definition and the `DiscoveryRegistrationCredential`.
+
+If a validation step fails, the presentation MUST be rejected.
+
+JWT credential and presentation encoding as specified by [VC-DATA-MODEL] MUST be applied.
+A clock skew of 5 seconds MAY be applied to `nbf` and `exp` claims.
+
+### 4.1 DiscoveryRegistrationCredential
```
gerardsn, line 167: add something about missing a proof / sharing proof with the VP

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,299 @@
...
+
+The optional `DiscoveryRegistrationCredential` is used to add additional parameters to the registration.
+It's a _holder_ credential, meaning it has been self issued. Fields under `credentialSubject` (except `id`) can be used by clients depending on the use case.
```
gerardsn, line 170: reference `holder credential` concept (Has it been removed from VC data model v2? I can't find it.)

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,299 @@
...
+- `exp` MUST be after `nbf`.
+- `aud` MUST contain the ID of the service.
+- the number of seconds between `nbf` and `exp` MUST NOT exceed `presentation_max_validity` (see Service Definition).
+- all credentials MUST have the same `credentialSubject.id`.
+- `exp` of the presentation MUST NOT be after the expiration date of the credentials.
+- the key used to sign the presentation MUST be owned by the credential subject (see 4.1):
+  the JWT `kid` header MUST reference an `assertionMethod` key from the subject's DID document.
+- The DID of the credential subject MUST be of a DID method defined in the service definition.
+- the presentation MAY contain a `DiscoveryRegistrationCredential` credential defined by the Nuts JSON-LD context (https://nuts.nl/credentials/v1).
+  If present, it MUST NOT be included in the next validation steps.
```
gerardsn, line 158: what if it is included in the presentation definition (next validation step)?

It makes sense to me that a use-case lists all the endpoints in the PD that are required for the use-case

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,299 @@
...
+    }
+  }
+]
+``
+
+## 7. Security Considerations
+
+## 7.1 Denial of Service
+
+Since the server is hosting a public endpoint, it should take the following measures to prevent denial-of-service attacks:
```
gerardsn, line 252: Should this apply to all nodes?
All nodes have the endpoint exposed and forward to what they consider the server. (Implementation detail of the nuts-node, so may need to be added to documentation on the node if valid.)

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,299 @@
...
+- limit the size of the presentations that can be uploaded.
+
+Clients should also take measures when the server has been compromised and is hosting malicious content:
+- limit the number of presentations to process.
+- limit the size of the presentations to process.
+- use streaming processing to prevent excessive memory usage.
+
+## 7.2 Trust
+
+Designers of services should build trust into the Presentation Definition of the server. This can be done by specifying specific issuers of credentials.
```
gerardsn, line 265:
```suggestion
Designers of services should write trust into the Presentation Definition of the Service Definition. This can be done by specifying specific issuers of credentials.
```

gerardsn (changes requested):
add [seed](https://github.com/nuts-foundation/nuts-node/issues/3463)

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,299 @@
...
+All presentations MUST conform to the Presentation Definition associated with the discovery service (see Service Definition).
+
+Presentations MUST be encoded in JWT format as string ([JWT]).
+
+The protocol identifies the server and client roles. A client can be a server and vice versa. The individual operations define what to do in certain cases.
+
+### 3.1 Registration
+
+Clients can publish a Verifiable Presentation to the server by sending it in an HTTP POST to the service endpoint.
+The HTTP request body MUST be a Verifiable Presentation (in JWT format). The content type of the request MUST be `application/json`.
```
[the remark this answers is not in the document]
woutslakhorst: we could drop this

rfc/rfc022-discovery-service.md
woutslakhorst: according to the spec, clients don't expose anything.

rfc/rfc022-discovery-service.md
woutslakhorst: added an unless.

rfc/rfc022-discovery-service.md
woutslakhorst: `holder claims`: https://www.w3.org/TR/vc-data-model-2.0/#presentations-including-holder-claims

woutslakhorst: changed to `self-asserted`

rfc/rfc022-discovery-service.md
```diff
@@ -0,0 +1,299 @@
...
+- it MUST contain a `retract_jti` JWT claim, containing the `jti` of the presentation to retract.
+- it MUST NOT contain any credentials.
+
+If a server receives a retraction that references an unknown presentation it MUST respond with a 400 Bad Request response.
+The response MUST be a JSON error response (§3.5) describing the error.
+
+Clients processing a retraction entry MUST remove the presentation indicated by `retract_jti`.
+The expiration of the retraction presentation MUST be equal to the expiration of the presentation to retract.
+
+### 3.5 Error responses
```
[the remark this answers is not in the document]
woutslakhorst: why specify the exact error string? There's nothing a user can do. All these errors are caused by a faulty client.

rfc/rfc022-discovery-service.md
woutslakhorst: should only be presentation

rfc/rfc022-discovery-service.md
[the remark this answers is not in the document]
gerardsn: IMO these are 3 different types of errors
- faulty client. action: nothing, or complain to your software provider (make issue for nuts-node)
- misconfigured client. action: users need to figure out who is at fault -> server/client (probably client)
- server dependency failure. action: monitor the issue and take action if it takes too long

The exact error message could be clearer, but I think these categories would help users debug any issues.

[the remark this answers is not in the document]
gerardsn: probably needs a general `ServerError` too

rfc/rfc022-discovery-service.md
gerardsn: I don't like magic and this also makes it harder to implement. It's unlikely a use-case will only contain the `authServerURL`, so I prefer the credential to be defined explicitly. If needed we can also add a flag to the API to add the `authServerURL` that defaults to true.

rfc/rfc022-discovery-service.md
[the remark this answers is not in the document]
woutslakhorst: in all cases an admin has to do something, not a user.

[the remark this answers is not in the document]
gerardsn: yes, but the error helps the admin figure out what to do. 

However, these suggestions were the result of o the original issue title in https://github.com/nuts-foundation/nuts-node/issues/3471. The changed scope of that issue makes this irrelevant

gerardsn (approved):

PR #268, opened 2023-12-01
update index and rfc graph

woutslakhorst:


.gitbook/assets/spec-relations.svg
gerardsn, line 1: so many lines...
this is getting impossible to read

gerardsn (approved):

PR #274, opened 2024-05-08
change tag to timestamp

woutslakhorst:


rfc/rfc022-discovery-service.md
```diff
@@ -48,16 +48,21 @@ All presentations MUST conform to the Presentation Definition associated with th
...
+
 ### 3.1 Registration
 
-Clients can publish a Verifiable Presentation to the list by sending it in an HTTP POST to the service endpoint.
+Clients can publish a Verifiable Presentation to the server by sending it in an HTTP POST to the service endpoint.
 The HTTP request body MUST be a Verifiable Presentation (in JWT format). The content type of the request MUST be `application/json`.
 
 The server MUST validate the Verifiable Presentation as specified in section 4. If the validation fails, it MUST return a 400 Bad Request response.
-If the validation succeeds, the Verifiable Presentation MUST be added to the list and the server MUST return a 201 Created response.
+If the validation succeeds, the Verifiable Presentation MUST be added to the internal list and the server MUST return a 201 Created response.
```
reinkrul, line 59: "internal list" is simply called "list" again below.

rfc/rfc022-discovery-service.md
```diff
@@ -48,16 +48,21 @@ All presentations MUST conform to the Presentation Definition associated with th
...
+If the Verifiable Presentation is already on the list, the server MUST return a 304 Not Modified response.
```
reinkrul, line 60: why?

How is equalness determined? VP ID is under control of the wallet (and thus might not be unique, although it should be), hash is complicated due to signature suite semantics or requires canonicalization.

rfc/rfc022-discovery-service.md
```diff
@@ -48,16 +48,21 @@ All presentations MUST conform to the Presentation Definition associated with th
...
+If the server is configured as client, it MUST send the presentation to the server it has configured.
```
reinkrul, line 61: "server is configured as client"?

A node is either client, or client and server (because client functionality is always enabled when a Service Definition is loaded)

rfc/rfc022-discovery-service.md
```diff
@@ -48,16 +48,21 @@ All presentations MUST conform to the Presentation Definition associated with th
...
 
 A credential subject (identified by `credentialSubject.id`) MUST NOT appear more than once on the list,
 so a new registration MUST replace the previous one from the same credential subject.
+If two presentations have the same issuance date, the server MUST keep the one with the highest byte value of the sha256 of the presentation.
```
reinkrul, line 65: I'd not use issuance date, since they can be out of order with the lamport timestamp. Furthermore, it's not required: if a lamport timestamp is used, it's impossible to have multiple VPs at the same time. Since each lamport timestamp increment = 1 event (aka VP).

With clock timestamps (e.g. unix time), duplicity would be a problem and you would need such a rule, but with lamport timestamps you don't (last write wins).

rfc/rfc022-discovery-service.md
```diff
@@ -74,40 +79,43 @@ The Verifiable Presentation MUST NOT be valid longer than the Verifiable Credent
 
 Clients MUST read the list by sending an HTTP GET to the server's service endpoint.
 The server MUST return a 200 OK response with a JSON object containing:
-- `entries` REQUIRED. MUST contain a JSON array with zero or more presentations.
-- `tag` REQUIRED. MUST be a JSON string referring to the last entry. The tag is an opaque value; the client SHOULD NOT derive any meaning from it.
+- `entries` REQUIRED. MUST contain a JSON object with a mapping of timestamp (as string) to presentation.
```
reinkrul, line 82: why is it a string? Also, this causes out-of-order map entries due to map serialization semantics, thus requires sorting.

Also, since the LC is monotonically increasing (thus, no gaps), you can simply return the first and last LCs.
Then also no sorting on the client side is required to make sure it processes the VPs in the ascending LC value order.

We might not need starting LC per se, but it could be by the client to determine out-of-syncness. And just for completeness

rfc/rfc022-discovery-service.md
```diff
@@ -74,40 +79,43 @@ The Verifiable Presentation MUST NOT be valid longer than the Verifiable Credent
...
+- `timestamp` REQUIRED. MUST be a JSON integer equal to the timestamp of the last presentation.
+
+The timestamp definition used here is a lamport clock, which is a monotonically increasing integer.
 
-The `tag` query parameter MAY be used by the client to request a delta next time it reads the list.
-If no `tag` query parameter is provided, the server MUST return the full list.
+The `timestamp` query parameter MAY be used by the client to request a delta next time it reads the list.
+The server SHOULD return the presentations with a timestamp greater than the provided value.
+The server MUST return the latest `timestamp` value based on its internal state.
```
reinkrul, line 89: "internal state" -> value of the last presentation on the list.

rfc/rfc022-discovery-service.md
```diff
@@ -74,40 +79,43 @@ The Verifiable Presentation MUST NOT be valid longer than the Verifiable Credent
...
+    "7": "ey1234..."
+  }
+  "timestamp": 7
 }
 ``
 
 Clients MUST validate each presentation in the list as specified in section 4.
 If a presentation is valid, the client uses it in its system. If a presentation is not valid, it MUST be rejected.
 If one or more presentations are not valid, it SHOULD NOT reject the other presentations.
+Clients SHOULD update the `timestamp` value based on the value of the return object.
```
reinkrul, line 111: "update the `timestamp` value" value of what?

rfc/rfc022-discovery-service.md
```diff
@@ -116,12 +124,14 @@ with the following additional requirements:
 
 - it MUST specify `RetractedVerifiablePresentation` as type, in addition to the `VerifiablePresentation`.
 - it MUST contain a `retract_jti` JWT claim, containing the `jti` of the presentation to retract.
+- it MUST contain a `retract_hash` JWT claim, containing the sha256 of the presentation to retract.
```
reinkrul, line 127: why? VP ID is under control of the signer, and should be unique?

rfc/rfc022-discovery-service.md
```diff
@@ -116,12 +124,14 @@ with the following additional requirements:
...
 - it MUST NOT contain any credentials.
 
 If a server receives a retraction that references an unknown presentation it MUST respond with a 400 Bad Request response.
 The response MUST be a JSON error response describing the error.
 
-Clients processing a retraction entry MUST remove the presentation indicated by `retract_jti`.
+Clients processing a retraction entry MUST remove the presentation indicated by `retract_jti` if the `retract_hash` matches the hash of the presentation to retract.
+The expiration of the retraction presentation MUST be equal to the expiration of the presentation to retract.
```
reinkrul, line 134: why?

reinkrul (changes requested):

rfc/rfc022-discovery-service.md
woutslakhorst: removed

rfc/rfc022-discovery-service.md
woutslakhorst: This is the registration part, the server will assign a LC to a submitted VP. A client can submit two VPs out of order/at the same time. The server will have to assign a LC value to both. Given the fact that they should come from the same party, using `nbf` should not lead to any problems.

rfc/rfc022-discovery-service.md
woutslakhorst: There could be holes: client submits two VPs, latest wins.
Using the key from the mapping as LC value for the client makes it deterministic and requires no sorting.

rfc/rfc022-discovery-service.md
woutslakhorst: changed to `Clients SHOULD use the `timestamp` value in the return object for their next call`

rfc/rfc022-discovery-service.md
woutslakhorst: obsolete

rfc/rfc022-discovery-service.md
woutslakhorst: An entry need not to be a mapping, it could also be a object with timestamp and vp fields.

rfc/rfc022-discovery-service.md
woutslakhorst: Added note that server must issue timestamps and last timestamp wins.

rfc/rfc022-discovery-service.md
```diff
@@ -108,7 +106,9 @@ Content-Type: application/json
 Clients MUST validate each presentation in the list as specified in section 4.
 If a presentation is valid, the client uses it in its system. If a presentation is not valid, it MUST be rejected.
 If one or more presentations are not valid, it SHOULD NOT reject the other presentations.
-Clients SHOULD update the `timestamp` value based on the value of the return object.
+If new presentations contains an already known credential subject (identified by `credentialSubject.id`),
```
reinkrul, line 109:
```suggestion
If new presentations contain an already known credential subject (identified by `credentialSubject.id`), 
```

reinkrul (approved):

PR #279, opened 2024-10-17
RFC022: Remove expiration time requirement on retractions

gerardsn:
Was there a reason behind this claim? Perhaps a remnant from when retractions didn't have their own VP type?

If the requirement stays it should be implemented


woutslakhorst:
Yes, list pruning. The specification only says expired VPs should be pruned. If you can add retractions that are invalid until the end of times, you can overload the server. Another implementation option would be that the server also prunes retractions when it prunes the registration. For that option, a hint should be added to the spec as well.

woutslakhorst (approved):

PR #281, opened 2024-11-05
Use a valid timestamp in the Organization Credential example

MartijnMuijsers:
73 is not a valid number of minutes ;p


gerardsn (approved):
Thanks Martijn!

PR #290, opened 2025-12-12
RFC023: Fixes

reinkrul:
- VC was missing `x5c#S256` header
- VC key ID was incorrect (`#1` vs `#0`)


stevenvegt (commented):
Should we also describe the contents of the header beside the example?

stevenvegt (approved):

